PR #4, opened 2022-04-14
locking parts of code

ilya-kozyrev:


alexeyinkin:
This PR introduces a separate widget for code with frozen blocks.

Its input is not raw text but a JSON that marks frozen blocks like this: https://github.com/akvelon/flutter-code-editor/blob/feature/6-lockingPartsOfCode/example/assets/settings/blockSettings.json
And it is a burden to convert examples to this format.

The representation is a `ListView` of separate `CodeField` editors, and those of read-only code are disabled. This disallows the following features:
- Selection of read-only blocks.
- Selection spanning from one editable block to another editable block through a read-only block.
- Future code folding if the folding block crosses the editors' border, because to each of the editors they contain un-opend or unterminated braces.

Due to these limitations we cannot use this approach.

PR #20, opened 2022-07-18
Implemented Read-Only Lines feature

miamihotline:
Implemented code split methods & read-only lines functionality

### DIFF
--- a/lib/src/code_field/models/code_model.dart
+++ b/lib/src/code_field/models/code_model.dart
@@ -1,0 +1,103 @@
+import 'package:flutter/material.dart';
+
+import 'code_line_model.dart';
+
+class CodeModel {
+  final String code;
+  late final List<CodeLineModel> lines;
+  late final List<CodeLineModel> readOnlyLines;
+
+  static final _cache = {};
+
+  factory CodeModel({required String code}) {
+    return _cache.putIfAbsent(code, () => CodeModel._(code));
+  }
+
+  CodeModel._(this.code) {
+    lines = splitTextToLines(code);
+    readOnlyLines = getReadOnlyLines(lines);
+  }
+
+  List<CodeLineModel> splitTextToLines(String text) {
+    int positionIndex = 0;
+    int lineCounter = 0;
+    List<String> codeLines = [];
+    List<CodeLineModel> lines = [];
+
+    if (text.isNotEmpty) {
+      codeLines = text.split('\n');
+
+      for (String line in codeLines) {
+        lineCounter++;
+        if (lines.isNotEmpty) {
+          positionIndex++;
+        }
+        lines.add(
+          CodeLineModel(
+            text: line,
+            lineNumber: lineCounter,
+            startIndex: positionIndex,
+            endIndex: positionIndex + line.length,
+            isReadOnly: line.contains('//readonly'),
+          ),
+        );
+        positionIndex += line.length;
+      }
+    }
+    return lines;
+  }
+
+  List<CodeLineModel> getReadOnlyLines(List<CodeLineModel> lines) {
+    List<CodeLineModel> readOnlyLines = [];
+
+    for (var line in lines) {
+      if (line.isReadOnly) {
+        readOnlyLines.add(line);
+      }
+    }
+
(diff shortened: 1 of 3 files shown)


lib/src/code_field/models/code_model.dart
```diff
@@ -0,0 +1,103 @@
+import 'package:flutter/material.dart';
+
+import 'code_line_model.dart';
+
+class CodeModel {
+  final String code;
+  late final List<CodeLineModel> lines;
+  late final List<CodeLineModel> readOnlyLines;
```
alexeyinkin, line 8: https://medium.com/flutter-senior/avoiding-late-variables-in-dart-abdb4c2b57e7

lib/src/code_field/models/code_model.dart
```diff
@@ -0,0 +1,103 @@
...
+
+  static final _cache = {};
```
alexeyinkin, line 10: A map from what to what?

lib/src/code_field/models/code_model.dart
```diff
@@ -0,0 +1,103 @@
...
+  factory CodeModel({required String code}) {
+    return _cache.putIfAbsent(code, () => CodeModel._(code));
+  }
+
+  CodeModel._(this.code) {
+    lines = splitTextToLines(code);
+    readOnlyLines = getReadOnlyLines(lines);
+  }
+
+  List<CodeLineModel> splitTextToLines(String text) {
```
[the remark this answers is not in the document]
alexeyinkin: Also everything should be private until proven otherwise.

lib/src/code_field/models/code_model.dart
```diff
@@ -0,0 +1,103 @@
...
+    int positionIndex = 0;
+    int lineCounter = 0;
+    List<String> codeLines = [];
```
alexeyinkin, line 24: If its use is narrow-scoped, declaration should be too.

lib/src/code_field/models/code_model.dart
```diff
@@ -0,0 +1,103 @@
...
+    List<CodeLineModel> lines = [];
+
+    if (text.isNotEmpty) {
```
alexeyinkin, line 27: This condition can be reversed to return something like `const Code.empty()`. This will unindent the code for non-empty case.

lib/src/code_field/models/code_model.dart
```diff
@@ -0,0 +1,103 @@
...
+      codeLines = text.split('\n');
+
+      for (String line in codeLines) {
+        lineCounter++;
+        if (lines.isNotEmpty) {
```
alexeyinkin, line 32: This can be done without special-casing some iterations to make code more clear and robust. If the first line's start position must be 1, `positionIndex` should be initialized with 1.

lib/src/code_field/models/code_model.dart
```diff
@@ -0,0 +1,103 @@
...
+        positionIndex += line.length;
+      }
+    }
+    return lines;
+  }
+
+  List<CodeLineModel> getReadOnlyLines(List<CodeLineModel> lines) {
+    List<CodeLineModel> readOnlyLines = [];
+
+    for (var line in lines) {
```
alexeyinkin, line 53: `final`

lib/src/code_field/models/code_model.dart
```diff
@@ -0,0 +1,103 @@
...
+            isReadOnly: line.contains('//readonly'),
+          ),
+        );
+        positionIndex += line.length;
+      }
+    }
+    return lines;
+  }
+
+  List<CodeLineModel> getReadOnlyLines(List<CodeLineModel> lines) {
```
alexeyinkin, line 50: `=> lines.where((line) => line.isReadOnly)` ?

lib/src/code_field/models/code_model.dart
alexeyinkin, line 32: This can be done without special-casing some iterations to make code more clear and robust.

alexeyinkin:
Done in #35 instead.

Issue #17, opened 2022-07-07
Fix false positive errors with default examples in Playground


alexeyinkin:
Our fork is meant to flag unterminated pair characters. It currently does it wrong and falsely flags unterminated braces in Java's `MinimalWordCount` and probably other examples.

PR #36, opened 2022-07-22
Disabled countingBrackets to fix false positive errors with default examples in Playground 

nausharipov:
The reason for the false positive errors is ' and " characters in words in multi-line comments

#17

### DIFF
--- a/lib/src/wip/getErrorsMap.dart
+++ b/lib/src/wip/getErrorsMap.dart
@@ -5,7 +5,6 @@ import 'package:highlight/languages/java.dart';
 import 'package:highlight/languages/python.dart';
 import 'package:highlight/languages/scala.dart';
 
-import 'language_syntax/brackets_counting.dart';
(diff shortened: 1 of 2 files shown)


alexeyinkin:
We talked on a call. Summary: this will not cover the case of
```
/*
 '
*/
```

Need to either employ `highlight` package's parsing result, or disable the feature.

lib/src/wip/getErrorsMap.dart
```diff
@@ -5,15 +5,17 @@ import 'package:highlight/languages/java.dart';
 import 'package:highlight/languages/python.dart';
 import 'package:highlight/languages/scala.dart';
 
-import 'language_syntax/brackets_counting.dart';
```
alexeyinkin, line 8: If you comment out the line below, it's better to also comment this out instead of deleting. That way we have less surprises when we uncomment the line below.

alexeyinkin (approved):

Issue #34, opened 2022-07-22
Derive use language-specific comments when parsing EOL comment tokens


alexeyinkin:
In
- #22
we are parsing both `//` and `#` comments for any language in `Code`.

We need to parse end-of-line comments that are specific to the language.

We need to support:
- Dart.
- Java.
- Go.
- Scala.
- Python.

For other languages, do not process end-of-line comments in `Code`.

PR #37, opened 2022-07-27
Replaced singleLineComments with getCommentSignByLanguage

nausharipov:
#34

### DIFF
--- a/lib/src/code/code.dart
+++ b/lib/src/code/code.dart
@@ -30,7 +30,7 @@ class Code {
     required String text,
     Mode? language,
   }) {
-    final singleLineComments = _getCommentByLanguage(language);
+    final singleLineComments = _getCommentsByLanguage(language);
     final lines = _textToCodeLines(text, singleLineComments);
     return Code._(text: text, lines: lines);
   }
@@ -81,7 +81,7 @@ class Code {
     return result;
   }
 
-  static List<String> _getCommentByLanguage(Mode? language) {
+  static List<String> _getCommentsByLanguage(Mode? language) {
     List<String>? singleLineComment = _singleLineComments[language];
     return singleLineComment ?? [];
   }
--- a/lib/src/code/code.dart
+++ b/lib/src/code/code.dart
@@ -91,7 +91,7 @@ class Code {
 
       return str
           .substring(commentIndex + singleLineComment.length)
-          .split(RegExp(r'\s+'));
+          .split(RegExp(r'\s+')); // Split by any whitespaces.
     } else {
       return [];
     }
--- a/lib/src/code/code.dart
+++ b/lib/src/code/code.dart
@@ -11,7 +11,16 @@ import 'code_line.dart';
 import 'text_range.dart';
 import 'tokens.dart';
 
-final List<Mode> slashCommentSignLanguages = [java, dart, go, scala];
+const _slashes = '//';
+const _hash = '#';
+
+final _singleLineComments = <Mode, List<String>>{
+  java: [_slashes],
+  dart: [_slashes],
+  go: [_slashes],
+  scala: [_slashes],
+  python: [_hash],
+};
 
 class Code {
   final String text;
@@ -19,14 +28,11 @@ class Code {
 
   factory Code({
     required String text,
-    required Mode? language,
+    Mode? language,
   }) {
-    final singleLineComment = _getCommentSignByLanguage(language);
-    final lines = _textToCodeLines(text, singleLineComment);
-    return Code._(
-      text: text,
-      lines: lines,
-    );
+    final singleLineComments = _getCommentByLanguage(language);
+    final lines = _textToCodeLines(text, singleLineComments);
+    return Code._(text: text, lines: lines);
   }
 
   const Code._({
@@ -38,7 +44,7 @@ class Code {
 
   static List<CodeLine> _textToCodeLines(
     String text,
-    String? singleLineComment,
+    List<String> singleLineComments,
   ) {
     final result = <CodeLine>[];
     final lines = text.split('\n');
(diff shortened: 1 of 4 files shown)


lib/src/code/code.dart
```diff
@@ -26,15 +38,15 @@ class Code {
 
   static List<CodeLine> _textToCodeLines(
     String text,
-    List<String> singleLineComments,
+    String? singleLineComment,
```
alexeyinkin, line 41: Some languages support multiple sequences for single line comments:
https://www.php.net/manual/en/language.basic-syntax.comments.php

So this should still be a list.

lib/src/code/code.dart
```diff
@@ -63,23 +75,26 @@ class Code {
     return result;
   }
 
-  static List<String> _getCommentWords(
-    String str,
-    List<String> singleLineComments,
-  ) {
-    for (final singleLineComment in singleLineComments) {
-      final commentIndex = str.indexOf(singleLineComment);
+  static String? _getCommentSignByLanguage(Mode? language) {
```
alexeyinkin, line 78: This is a sequence and not a sign.

lib/src/code/code.dart
```diff
@@ -1,20 +1,32 @@
...
+import 'package:highlight/languages/go.dart';
+import 'package:highlight/languages/java.dart';
+import 'package:highlight/languages/python.dart';
+import 'package:highlight/languages/scala.dart';
 
 import 'code_line.dart';
 import 'text_range.dart';
 import 'tokens.dart';
 
+final List<Mode> slashCommentSignLanguages = [java, dart, go, scala];
```
alexeyinkin, line 14: The cleanest way I can think of is
```dart
const _slashes = '//';
const _hash = '#';

const _singleLineComments = <String, List<String>>{
  {java: [_slashes]},
  {python: [_hash]},
  // ...
};
```

Then just pick from the map without `if`s.

lib/src/code_field/code_controller.dart
```diff
@@ -97,7 +94,7 @@ class CodeController extends TextEditingController {
 
   CodeController({
     String? text,
-    Mode? language,
+    required Mode? language,
```
alexeyinkin, line 97: `language` should not be required. It breaks backwards compatibility. If we want others to migrate to our editor, we must support as much use cases as possible.

lib/src/code_field/code_controller.dart
```diff
@@ -317,10 +313,7 @@ class CodeController extends TextEditingController {
         return;
       }
 
-      _lastCode = Code(
-        text: newValue.text,
-        singleLineComments: _singleLineComments,
```
alexeyinkin, line 322: Only this line should change. Preserving lines is a virtue.

lib/src/code/code.dart
```diff
@@ -1,20 +1,32 @@
...
+
 class Code {
   final String text;
   final List<CodeLine> lines;
 
   factory Code({
     required String text,
-    required List<String> singleLineComments,
+    required Mode? language,
```
alexeyinkin, line 22: Should not be required.

lib/src/code/code.dart
```diff
@@ -1,20 +1,32 @@
...
   }) {
-    final lines = _textToCodeLines(text, singleLineComments);
-    return Code._(text: text, lines: lines);
```
alexeyinkin, line 17: This line did not have to change.

lib/src/code/code.dart
```diff
@@ -63,23 +75,26 @@ class Code {
...
-      if (commentIndex == -1) {
-        continue;
-      }
+  static List<String> _getCommentWords(String str, String? singleLineComment) {
+    if (singleLineComment != null && str.contains(singleLineComment)) {
+      final commentIndex = str.indexOf(singleLineComment);
 
       return str
           .substring(commentIndex + singleLineComment.length)
           .split(RegExp(r'\s+')); // Split by any whitespaces.
```
alexeyinkin, line 94: You deleted this comment and then re-created it. So in git history it is now yours and not mine. History is very important when looking for what introduced errors. So when you revert the changes, you should squash the commits and publish a new branch.

test/src/code/code_test.dart
```diff
@@ -163,8 +165,8 @@ void main() {
             readonly
             readonly //readonly
             readonly // a readonly b'''
-            '\n\n'
-            '''
+              '\n\n'
+              '''
```
alexeyinkin, line 169: ?

test/src/code/code_test.dart
```diff
@@ -4,9 +4,11 @@ import 'package:code_text_field/src/code/code.dart';
 import 'package:code_text_field/src/code/code_line.dart';
 import 'package:collection/collection.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:highlight/highlight.dart';
+import 'package:highlight/languages/angelscript.dart';
+import 'package:highlight/languages/java.dart';
 
-const singleLineComments = ['//', '#'];
-
+final Mode language = java;
```
alexeyinkin, line 11: We prefer inferred typing as it is more readable: `final _language = java`.
Please also make it private.
Please also make `text` private.

test/src/code/code_test.dart
nausharipov: My formatter works differently. I'll look into it.

test/src/code_field/code_controller_test.dart
```diff
@@ -16,12 +18,17 @@ abc
 ro//readonly
 ''';
 
+final Mode language = java;
```
alexeyinkin, line 21: Private and without type, please.

alexeyinkin (changes requested):
Thank you. Here are the comments.

PR #38, opened 2022-07-28
Language-specific comments when parsing EOL comment tokens

nausharipov:
#34

### DIFF
--- a/lib/src/code/code.dart
+++ b/lib/src/code/code.dart
@@ -1,9 +1,26 @@
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
+import 'package:highlight/highlight.dart';
+import 'package:highlight/languages/dart.dart';
+import 'package:highlight/languages/go.dart';
+import 'package:highlight/languages/java.dart';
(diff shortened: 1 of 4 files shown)


lib/src/code/code.dart
```diff
@@ -1,18 +1,36 @@
...
 class Code {
   final String text;
   final List<CodeLine> lines;
 
   factory Code({
     required String text,
-    required List<String> singleLineComments,
+    Mode? language,
   }) {
+    final singleLineComments = _getCommentsByLanguage(language);
```
alexeyinkin, line 33: If it's so simple, we don't need the function anymore.

test/src/code/code_test.dart
```diff
@@ -190,5 +191,13 @@ void main() {
         }
       }
     });
+    test(
```
alexeyinkin, line 194: Separate tests by a blank line please.

alexeyinkin (changes requested):
Thanks.

alexeyinkin (approved):

PR #41, opened 2022-08-04
Use highlight to parse comments when possible

alexeyinkin:


lib/src/code/code.dart
```diff
@@ -75,29 +79,19 @@ class Code {
...
       );
+
       charIndex += line.length + 1;
+      lineIndex++;
     }
 
     return result;
   }
 
   static List<String> _getCommentWords(
```
nausharipov, line 90: if this function is so simple, is it needed?

lib/src/single_line_comments/parser/single_line_comments.dart
```diff
@@ -0,0 +1,21 @@
...
+import 'package:highlight/languages/dart.dart';
+import 'package:highlight/languages/go.dart';
+import 'package:highlight/languages/java.dart';
+import 'package:highlight/languages/python.dart';
+import 'package:highlight/languages/scala.dart';
+
+class SingleLineComments {
+  const SingleLineComments._();
+
+  static final byMode = <Mode, List<String>>{
```
nausharipov, line 11: why not byLanguage?

alexeyinkin: The users would have to guess if it is the `Mode` object or a string alias.

lib/src/code/code.dart
alexeyinkin: We may want to strip commas or something like that later. This is also a different level of abstraction. A function should not mix different levels of abstractions.

test/src/single_line_comment/parser/highlight_single_line_comment_parser_test.dart
```diff
@@ -0,0 +1,311 @@
...
+        text: '''
+val str4 = """
+           // not a comment
+           """;
+''',
+        isLanguageLost: true,
+      ),
+    ];
+
+    for (final one in data) {
```
nausharipov, line 276: is there a better word than "one"? (case, item, etc.)

test/src/code/code_test.dart
```diff
@@ -49,7 +49,7 @@ void main() {
       }
     });
 
-    test('parses lines', () {
+    test('Parses lines', () {
```
nausharipov, line 52: other test descriptions are not capitalized. Is there a reason for this specific description to be capitalized in this PR?

test/src/single_line_comment/parser/highlight_single_line_comment_parser_test.dart
```diff
@@ -0,0 +1,311 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:highlight/highlight_core.dart';
+import 'package:highlight/languages/go.dart';
+import 'package:highlight/languages/java.dart';
+import 'package:highlight/languages/python.dart';
+import 'package:highlight/languages/scala.dart';
+
+void main() {
+  test('HighlightSingleLineCommentParser', () {
+    final data = [
```
nausharipov, line 13: this array takes 250 lines. Would it be better to separate "data" into a separate file?

test/src/single_line_comment/parser/single_line_comment_parser_test.dart
```diff
@@ -0,0 +1,87 @@
...
+              SingleLineComments.byMode[one.language] ?? [],
+        );
+
+        expect(result.runtimeType, one.parserType);
+      }
+    });
+  });
+}
+
+class _Data {
```
nausharipov, line 75: isn't _Case or _TestCase a better name for such classes?

test/src/code/code_test.dart
alexeyinkin: https://medium.com/@alexey.inkin/how-to-name-and-group-tests-in-a-dart-file-553e194bd8af

test/src/single_line_comment/parser/highlight_single_line_comment_parser_test.dart
alexeyinkin: +

test/src/single_line_comment/parser/text_single_line_comment_parser_test.dart
```diff
@@ -0,0 +1,59 @@
+import 'package:code_text_field/src/single_line_comments/parser/text_single_line_comment_parser.dart';
+import 'package:code_text_field/src/single_line_comments/single_line_comment.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('TextSingleLineCommentParser', () {
+    test('grabs everything at sequence and on', () {
```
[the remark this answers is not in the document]
alexeyinkin: This one is quite common.
image
https://www.merriam-webster.com/dictionary/on

test/src/single_line_comment/parser/single_line_comment_parser_test.dart
alexeyinkin: `+`

test/src/single_line_comment/parser/highlight_single_line_comment_parser_test.dart
alexeyinkin: 1. It is no reused and will likely never by.
2. That would make the test case class public.
3. This array and the test it's in have many common reasons to change quite separate from other reasons affecting other files, so they are natural together.

test/src/single_line_comment/parser/text_single_line_comment_parser_test.dart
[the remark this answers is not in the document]
nausharipov: yeah, but I don't quite understand the meaning of 'at sequence and on'. I guess it has its own nuance.

PR #51, opened 2022-08-08
Use named sections for read-only code

### LINKED ISSUE #39: Read-only by named sections
In #22, we added read-only blocks by new line blocks. Now we need to allow read-only blocks by named sections like:
```
editable
[START section_name]
readonly
readonly
[END section_name]
editable
```

This should work if the given `section_name` is passed as a read-only section. Multiple such section names are allowed to be read-only.

alexeyinkin:
Resolves #39.

### DIFF
--- a/lib/src/code/code.dart
+++ b/lib/src/code/code.dart
@@ -2,8 +2,11 @@ import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:highlight/highlight_core.dart';
 
+import '../named_sections/named_section.dart';
+import '../named_sections/parsers/abstract.dart';
 import '../single_line_comments/parser/single_line_comment_parser.dart';
 import '../single_line_comments/parser/single_line_comments.dart';
+import '../single_line_comments/single_line_comment.dart';
 import 'code_line.dart';
 import 'text_range.dart';
 import 'tokens.dart';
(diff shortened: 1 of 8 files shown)


lib/src/code/code.dart
```diff
@@ -2,55 +2,84 @@ import 'package:collection/collection.dart';
...
+    );
+
+    final sections = namedSectionParser?.parse(
+          singleLineComments: commentParser.comments,
+          lineCount: lines.length,
+        ) ??
+        const [];
+    final sectionsMap = {for (final s in sections) s.name: s};
+
+    _applyNamedSectionsToLines(
```
nausharipov, line 48: can this be a pure function with no side effects?

test/src/code/code_test.dart
```diff
@@ -11,7 +12,7 @@ final _language = java;
 const _text = '''
 1 Lorem ipsum dolor sit amet,
 2 consectetur adipiscing elit,
-3 sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+3 sed do eiusmod tempor incididunt ut labore et dolore magna.
```
nausharipov, line 15: can't "aliqua" be added to the line #73 instead?

lib/src/code/code.dart
alexeyinkin: This will run on each change in the editor, so I don't want to slow it down.

For the same reason I also think of making a `CodeLine` mutable to just change its `isReadOnly` in there, but then we can't override `hashCode` and `operator ==`, so I have not yet decided on this.

Anyway tests are there to compensate for the risks of side effects.

test/src/code/code_test.dart
alexeyinkin: `+` Good idea, less changes, thanks.

PR #55, opened 2022-08-26
Dictionary-based autocomplete

### LINKED ISSUE #23: Implement auto-completion of symbols based on dictionary
We need to be able to provide to the flutter-code-editor library an information about the key symbols that will be used for autocompletion, when user types code.

alexeyinkin:
Resolves #23


test/src/code_field/text_editing_value_test.dart
```diff
@@ -0,0 +1,381 @@
...
+      for (final example in examples) {
+        final result = example.value.wordAtCursor;
+
+        expect(result, example.expected, reason: example.name);
+      }
+    });
+
+    test('wordToCursor', () {
+      const examples = [
+        //
```
nausharipov, line 255: Are these empty comments a part of examples? If so, is it clear for others?

nausharipov (approved):

test/src/code_field/text_editing_value_test.dart
alexeyinkin: This forces the Dart formatter to preserve spaces between examples.
It is not clear for others.

PR #57, opened 2022-09-05
Add HighlightFoldableBlockParser, add tests for Java

alexeyinkin:
Addresses #44

### DIFF
--- a/lib/src/folding/parsers/abstract.dart
+++ b/lib/src/folding/parsers/abstract.dart
@@ -1,0 +1,244 @@
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+
+import '../foldable_block.dart';
+import '../foldable_block_type.dart';
+import '../invalid_foldable_block.dart';
+import 'line_semantics.dart';
+
+/// A base class for parsers that go through some representation
+/// of code to parse blocks from it.
+///
+/// This object accumulates data in its fields as the parsing progresses.
+class AbstractFoldableBlockParser {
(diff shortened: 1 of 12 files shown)


lib/src/folding/parsers/abstract.dart
```diff
@@ -0,0 +1,244 @@
...
+          startLine: first,
+          endLine: last!,
+          type: FoldableBlockType.singleLineComment,
+        ),
+      );
+    }
+
+    int i = lineIndex;
+
+    for (final line in _linesWithSemantics.reversed) {
```
nausharipov, line 177: endCommentSequenceIfAny, endImportSequenceIfAny and other similar functions seem too big. Can they be broken down into smaller ones?

lib/src/folding/parsers/line_semantics.dart
```diff
@@ -0,0 +1,34 @@
+import '../../highlight/keyword_semantics.dart';
+
+/// The semantics of a line from the perspective of a sequence
+/// of similar lines that can from a foldable block.
```
nausharipov, line 4: a typo: "that can from".

alexeyinkin: `+`

lib/src/folding/parsers/abstract.dart
alexeyinkin: I can't think of a way that would not also make it slower.

This will be called on every character change for every line in the code, so it may be critical, although the performance is not tested.

This code complexity could be tolerated for code covered well.

PR #60, opened 2022-09-07
Highlight invalid blocks

### LINKED ISSUE #59: Highlight invalid blocks of pair characters

alexeyinkin:
Resolves #59


lib/src/gutter/gutter.dart
```diff
@@ -0,0 +1,57 @@
...
+    final tableRows = [
+      for (int i = 1; i <= code.lines.length; i++)
+        TableRow(
+          children: [
+            Text(
+              '$i',
+              style: style.textStyle,
+              textAlign: style.textAlign,
+            ),
+            const SizedBox(),
```
nausharipov, line 28: What is this SizedBox for?

lib/src/gutter/gutter.dart
```diff
@@ -0,0 +1,57 @@
...
+          ],
+        ),
+    ];
+
+    for (final issue in code.issues) {
+      tableRows[issue.line].children![1] = Container(
```
nausharipov, line 34: Will the meaning of the magic number 1 be clear to others?

nausharipov (approved):

lib/src/gutter/gutter.dart
alexeyinkin: To reserve space. Its width comes from `_foldingColumnWidth`.
Without it, the entire width is taken by line numbers.

PR #63, opened 2022-09-09
Move up the autocomplete popup if it is cropped

### LINKED ISSUE #19: The suggestion box is cropped when at the bottom
image

nausharipov:
Resolves #19

### DIFF
--- a/lib/src/code_field/code_field.dart
+++ b/lib/src/code_field/code_field.dart
@@ -7,6 +7,7 @@ import 'package:linked_scroll_controller/linked_scroll_controller.dart';
 import '../code_theme/code_theme.dart';
 import '../gutter/gutter.dart';
 import '../line_numbers/line_number_style.dart';
+import '../sizes.dart';
 import '../wip/autocomplete/popup.dart';
 import 'code_controller.dart';
 
@@ -89,8 +90,8 @@ class CodeFieldState extends State<CodeField> {
   ScrollController? _horizontalCodeScroll;
   final _codeFieldKey = GlobalKey();
 
-  double cursorX = 0;
-  double cursorY = 0;
+  Offset _preferredPopupOffset = Offset.zero;
+  bool _isPopupCropped = false;
   double painterWidth = 0;
   double painterHeight = 0;
 
@@ -117,8 +118,8 @@ class CodeFieldState extends State<CodeField> {
     _focusNode!.attach(context, onKeyEvent: _onKeyEvent);
 
     WidgetsBinding.instance.addPostFrameCallback((_) {
-      double width = _codeFieldKey.currentContext!.size!.width;
-      double height = _codeFieldKey.currentContext!.size!.height;
+      final double width = _codeFieldKey.currentContext!.size!.width;
+      final double height = _codeFieldKey.currentContext!.size!.height;
       windowSize = Size(width, height);
     });
     _onTextChanged();
@@ -148,8 +149,8 @@ class CodeFieldState extends State<CodeField> {
         // so check first.
         final context = _codeFieldKey.currentContext;
         if (context != null) {
-          double width = context.size!.width;
-          double height = context.size!.height;
+          final double width = context.size!.width;
+          final double height = context.size!.height;
           windowSize = Size(width, height);
         }
       });
@@ -306,17 +307,16 @@ class CodeFieldState extends State<CodeField> {
             child: Stack(
               children: [
                 editingField,
-                widget.controller.popupController.isPopupShown
-                    ? Popup(
-                        row: cursorY,
-                        column: cursorX,
-                        controller: widget.controller.popupController,
-                        editingWindowSize: windowSize,
-                        style: textStyle,
-                        backgroundColor: backgroundCol,
-                        parentFocusNode: _focusNode!,
-                      )
-                    : Container(),
+                if (widget.controller.popupController.isPopupShown)
+                  Popup(
+                    preferredOffset: _preferredPopupOffset,
+                    isPopupCropped: _isPopupCropped,
+                    controller: widget.controller.popupController,
+                    editingWindowSize: windowSize,
+                    style: textStyle,
+                    backgroundColor: backgroundCol,
+                    parentFocusNode: _focusNode!,
+                  ),
               ],
             ),
           ),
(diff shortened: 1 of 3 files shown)


alexeyinkin:
Also it would be nice to fix the horizontal offset once we are here:

image
image

This is how it is expected to work in IDEs:
image

nausharipov:
IDEs work differently: 
Screen Shot 2022-09-12 at 12 51 49

Do we need to follow Android Studio's preferences?

alexeyinkin:
> Do we need to follow Android Studio's preferences?

No, but let's make it at least consistent.

With `cl`, the tip is 1 char to the left:
image

With `cla`, the tip is aligned to the input:
image

With `clas`, the tip tip is 1 char to the right:
image

lib/src/code_field/code_field.dart
```diff
@@ -393,8 +397,17 @@ class CodeFieldState extends State<CodeField> {
...
-      painterWidth = painter.width;
-      painterHeight = painter.height;
+      _positionPopup(caretHeight);
     });
   }
+
+  void _positionPopup(double caretHeight) {
+    if (cursorY + Sizes.autocompletePopupMaxHeight > windowSize.height) {
+      cursorY = cursorY -
+          (Sizes.autocompletePopupMaxHeight + caretHeight + Sizes.caretPadding);
```
alexeyinkin, line 407: What is `cursorY` defined to store?

lib/src/wip/autocomplete/popup.dart
```diff
@@ -38,7 +41,7 @@ class _PopupState extends State<Popup> {
   void initState() {
     widget.controller.addListener(rebuild);
     this.width = 300;
```
alexeyinkin, line 43: Paired variables like `width` and `height` should be processed alike. If you drop `this.` on one, please do so on the other too even though it is beyond the min required change.

lib/src/wip/autocomplete/popup.dart
```diff
@@ -50,13 +53,13 @@ class _PopupState extends State<Popup> {
...
   Widget build(BuildContext context) {
-    return Padding(
-      padding: EdgeInsets.only(
-        left: min(widget.column, widget.editingWindowSize.width - width),
-        top: widget.row,
-      ),
-      child: ConstrainedBox(
+    return Positioned(
+      left: min(widget.column, widget.editingWindowSize.width - width),
+      top: widget.row,
```
alexeyinkin, line 58: I moved this file to the `wip` dir because it is messy and needs refactoring, so please do not rely on anything here. One messy thing is that `row` and `column` here actually mean `y` and `x` in pixels although the names imply that these are characters.

Please fix such things as you notice them. If unsure, suggest such fixes first.

lib/src/wip/autocomplete/popup.dart
```diff
@@ -50,13 +53,13 @@ class _PopupState extends State<Popup> {
...
   @override
   Widget build(BuildContext context) {
-    return Padding(
-      padding: EdgeInsets.only(
-        left: min(widget.column, widget.editingWindowSize.width - width),
-        top: widget.row,
-      ),
-      child: ConstrainedBox(
+    return Positioned(
+      left: min(widget.column, widget.editingWindowSize.width - width),
```
alexeyinkin, line 57: The problem here is that this blurs the responsibility of positioning. The positioning should be one in one of these ways:

1. The outer widget determines the exact position and only passes the offset, and not alignment or window size.
2. The outer widget passes the caret position and the window size and lets this widget decide on the rest.

The 2nd one follows the SRP closer because the field has less reasons to change.

alexeyinkin: Also offsets should normally not be passed as separate coördinates because it means twice the job to handle. I suggest you make a single property for `x` and `y`. It can be `offset` if you let the outer widget position the popup or `preferredOffset` if you delegate this to the popup. Flutter has the `Offset` class for this.

alexeyinkin (changes requested):
Thanks.

lib/src/wip/autocomplete/popup.dart
nausharipov: Done.

lib/src/code_field/code_field.dart
alexeyinkin: If we use the preferred offset, then the host widget should only calculate the offset for the default popup position (which is below the cursor) and let the popup widget handle the rest. The host should not decide on the alignment for it. Only then is the SRP observed.

The popup widget will also need the height of the line if it decides to flip.

lib/src/code_field/code_field.dart
```diff
@@ -89,8 +90,8 @@ class CodeFieldState extends State<CodeField> {
   ScrollController? _horizontalCodeScroll;
   final _codeFieldKey = GlobalKey();
 
-  double cursorX = 0;
-  double cursorY = 0;
+  double popupLeftOffset = 0;
+  double popupTopOffset = 0;
```
alexeyinkin, line 94: If this widget calculates the offset once and does not adjust it later, then this can also be a single `Offset preferredPopupOffset` object.

alexeyinkin (changes requested):
Thanks.

lib/src/code_field/code_field.dart
```diff
@@ -56,7 +56,6 @@ class CodeField extends StatefulWidget {
   final bool lineNumbers;
 
   const CodeField({
-    Key? key,
```
alexeyinkin, line 58: This package aims to be a drop-in replacement for https://pub.dev/packages/code_text_field
so we can get their user base when we go public.

For this reason, we must preserve as much of its public API as we can.

So we do need `key` here.

Moreover, `key` is very much desired for any public package because you never know how it will be used. Our rule to not require keys is limited to our job for Apache because it is not reused much. For this public package we will eventually raise the QA standard to this: https://github.com/alexeyinkin/dart-total-lints/blob/main/lib/package_2.18.yaml

And we also must keep the key argument like this and not replace it with `super.key` for less diff against https://github.com/BertrandBev/code_field/blob/master/lib/src/code_field/code_field.dart
Until we go public, this has higher priority than fixing linter issues.

lib/src/code_field/code_field.dart
```diff
@@ -198,7 +195,7 @@ class CodeFieldState extends State<CodeField> {
               child: Text(longestLine, style: textStyle),
             ), // Add extra padding
           ),
-          widget.expands ? Expanded(child: codeField) : codeField,
+          if (widget.expands) Expanded(child: codeField) else codeField,
```
alexeyinkin, line 198: Please revert this for less diff with the original. This way it is easier to check if the original editor has any new features we need to grab to ourselves.

alexeyinkin (changes requested):
Thanks.

lib/src/code_field/code_field.dart
```diff
@@ -325,32 +325,63 @@ class CodeFieldState extends State<CodeField> {
...
-    TextPainter painter = TextPainter(
+  void _updatePopupOffset(String text) {
+    final TextPainter textPainter = _getTextPainter(text);
+    final caretHeight = _getCaretHeight(textPainter);
+
+    final double normalLeftOffset = _getPopupLeftOffset(textPainter);
+    final double normalTopOffset = _getPopupTopOffset(textPainter, caretHeight);
+    final double flippedLeftOffset =
+        // TODO(nausharipov): find where 100 comes from
+        windowSize.width - Sizes.autocompletePopupMaxWidth - 100;
```
alexeyinkin, line 336: Should not the left offset be the same for normal and flipped popups?

nausharipov: Screen Shot 2022-09-13 at 15 24 13
Screen Shot 2022-09-13 at 15 23 33

lib/src/wip/autocomplete/popup.dart
```diff
@@ -27,18 +26,24 @@ class Popup extends StatefulWidget {
...
+  static const double _maxHeight = Sizes.autocompletePopupMaxHeight;
+  bool _verticalOverflow = false;
+  bool _horizontalOverflow = false;
 
   @override
   void initState() {
     widget.controller.addListener(rebuild);
-    this.width = 300;
-    this.height = 100;
+    _verticalOverflow =
```
alexeyinkin, line 41: These should not be stored because they can change after `initState()`. Make them local variables in `build()`.

lib/src/code_field/code_field.dart
nausharipov: 1) If it is the same as normal.
2) If it is limited.

lib/src/wip/autocomplete/popup.dart
```diff
@@ -50,53 +55,47 @@ class _PopupState extends State<Popup> {
...
+      top: _verticalOverflow ? widget.flippedOffset.dy : widget.normalOffset.dy,
+      child: Container(
+        alignment:
+            _verticalOverflow ? Alignment.bottomCenter : Alignment.topCenter,
+        constraints: const BoxConstraints(
+          maxHeight: _maxHeight,
+          maxWidth: _maxWidth,
+        ),
+        // DecoratedBox is rendered differently.
+        // ignore_for_file: use_decorated_box
```
alexeyinkin, line 71: Try not to ignore anything for the entire file. Use just `ignore: ` for this line.

lib/src/wip/autocomplete/popup.dart
```diff
@@ -50,53 +55,47 @@ class _PopupState extends State<Popup> {
...
+          : widget.normalOffset.dx,
+      top: _verticalOverflow ? widget.flippedOffset.dy : widget.normalOffset.dy,
+      child: Container(
+        alignment:
+            _verticalOverflow ? Alignment.bottomCenter : Alignment.topCenter,
+        constraints: const BoxConstraints(
+          maxHeight: _maxHeight,
+          maxWidth: _maxWidth,
+        ),
+        // DecoratedBox is rendered differently.
```
alexeyinkin, line 70: That exactly is the problem with `DecoratedBox`? Can you describe in the comment?

Issue #64, opened 2022-09-13
Hidden comments generate foldable blocks


alexeyinkin:
Dmitry Repin
  [5:08 PM]

```
(https://akvelon.slack.com/archives/D041YQ1E1B4/p1662988084068089)
FoldableBlock(startLine: 0, endLine: 6, type: FoldableBlockType.braces)
FoldableBlock(startLine: 1, endLine: 2, type: FoldableBlockType.braces)
FoldableBlock(startLine: 2, endLine: 3, type: FoldableBlockType.singleLineComment)
FoldableBlock(startLine: 4, endLine: 5, type: FoldableBlockType.braces)
###
class MyClass {
  void readOnlyMethod() {// [START section1]
  }// [END section1]
  // [START section2]
  void method() {
  }// [END section2]
}
```

PR #65, opened 2022-09-13
issue 64 hidden comments dont generate foldable blocks

Malarg:


Malarg:
resolves #64

lib/src/folding/parsers/highlight.dart
```diff
@@ -2,10 +2,10 @@ import 'package:charcode/ascii.dart';
 import 'package:collection/collection.dart';
 import 'package:highlight/highlight_core.dart';
 
+import '../../../flutter_code_editor.dart';
```
alexeyinkin, line 5: This file imports everything this package exports.
I try to only import the specific files we need but it is not a conscious decision.
Can you help me list the benefits of each pattern?

lib/src/folding/parsers/highlight.dart
```diff
@@ -66,6 +66,19 @@ class HighlightFoldableBlockParser extends AbstractFoldableBlockParser {
   void _processComment(Node node) {
     final newlineCount = node.getNewlineCount();
 
+    if (_foundNonWhitespace) {
+      return;
+    }
+
+    if (node.children?.any(
+          (subnode) => BracketsStartEndNamedSectionParser.isNamedSectionTag(
```
alexeyinkin, line 74: There may be other types of hidden comments, so we should not rely on `isNamedSectionTag()` here.

In `Code` constructor, we have `serviceComments` as the source of truth for which comments are hidden, looks like we need to pass it to `HighlightFoldableBlockParser.parse()`.

Next is the task to match nodes with the comments to see if a node's comment is a service comment.

alexeyinkin (changes requested):
Thanks.

lib/src/folding/parsers/highlight.dart
Malarg: fixed

lib/src/folding/parsers/highlight.dart
Malarg: Unfortunately, I don't have a strong opinion here either. But I also think, that importing specific files is good practice, because this should make it easier for the compiler. Fixed here

lib/src/single_line_comments/single_line_comment.dart
```diff
@@ -14,11 +14,15 @@ class SingleLineComment {
   /// The comment text with the characters signifying the comment.
   final String outerContent;
 
+  //source, which provided comment to parse
```
alexeyinkin, line 17: 1. Doc comments should start with `///`. It's only then that they are compiled into the package documentation and are shown in IDE tips.

2. Every example in the guide has a single space after `//` or `///`, starts with a capital letter, and ends with a period: https://dart.dev/guides/language/effective-dart/documentation

3. In English, there should be no comma here.

4. According to the guide, we should PREFER using this when referring to the current object, like "this comment": https://dart.dev/guides/language/effective-dart/documentation#prefer-using-this-instead-of-the-to-refer-to-a-members-instance

5. "Source" is an unclear noun that leaves questions. I would have put it like this:

```dart
/// The object this comment was parsed from, if any.
///
/// The object class is specific to the parser.
```

test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -177,14 +180,57 @@ import java.lang.Exception;''',
           ],
           expectedInvalid: const [],
         ),
+        _Example(
+          'Java. Comment after close brace',
```
alexeyinkin, line 184: 'a closing'

test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -177,14 +180,57 @@ import java.lang.Exception;''',
...
+  // comment
+}''',
+          mode: java,
+          expected: const [
+            _FB(startLine: 0, endLine: 4, type: _T.braces),
+            _FB(startLine: 1, endLine: 2, type: _T.braces),
+          ],
+          expectedInvalid: const [],
+          namedSectionParser: const BracketsStartEndNamedSectionParser(),
+        ),
```
alexeyinkin, line 198: Please add empty lines between examples for better separation.

test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -177,14 +180,57 @@ import java.lang.Exception;''',
...
+
+        final commentParser = SingleLineCommentParser.parseHighlighted(
+          text: example.code,
+          highlighted: highlighted,
+          singleLineCommentSequences: sequences,
+        );
+
+        final serviceComments = ServiceCommentFilter.filter(
+          commentParser.comments,
+          namedSectionParser: example.namedSectionParser,
```
alexeyinkin, line 230: To close match real usage, I suggest we use a named section parser on all examples. If doing so, they either should be in each example, or `BracketStartEndNamedSectionParser` can be hardcoded here since it is the only one for now.

lib/src/code/code.dart
```diff
@@ -46,17 +47,12 @@ class Code {
...
       namedSectionParser: namedSectionParser,
     );
 
-    final lines = _textToCodeLines(
-      text: text,
-      highlighted: highlighted,
-      language: language,
-      commentsByLines: commentParser.getCommentsByLines(),
-    );
+    final serviceCommentsNodesSet = serviceComments.map((e) => e.source).toSet();
```
alexeyinkin, line 55: If this is duplicated, is it better to make an extension method?

```dart
extension SingleLineCommentIterable on Iterable<SingleLineComment> {
  Set<Object?> get sources => {... map((e) => e.source)};
}
```

This can be in `single_line_comment.dart`.

alexeyinkin (changes requested):
Thanks.

lib/src/single_line_comments/single_line_comment.dart
Malarg: Thank you

test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -177,14 +180,57 @@ import java.lang.Exception;''',
...
+        );
+
+        parser.parse(highlighted, serviceComments.map((e) => e.source).toSet());
```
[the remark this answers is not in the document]
Malarg: Nice feature, I did not know about it

Malarg:
fixed all comments

lib/src/single_line_comments/single_line_comment.dart
```diff
@@ -77,3 +79,7 @@ class SingleLineComment {
     throw Exception('$outerContent does not start with any of $sequences');
   }
 }
+
+extension SingleLineCommentIterable on Iterable<SingleLineComment> {
+  Set<Object?> get sources => {... map((e) => e.source)};
```
alexeyinkin, line 84: Please use auto-formatter on all files. This line is not formatted.

alexeyinkin (approved):

PR #66, opened 2022-09-15
Tab symbol is replaced with spaces on fullText set (#62)

### LINKED ISSUE #62: Replace tabs with spaces when loading the code
There was an issue in Beam Playground that suggests that tab symbols are not handled correctly in the editor:
https://issues.apache.org/jira/browse/BEAM-13446

Check this and fix if needed.

nausharipov:
Resolves #62

### DIFF
--- a/lib/src/code_field/code_controller.dart
+++ b/lib/src/code_field/code_controller.dart
@@ -16,6 +16,7 @@ import '../code_modifiers/tab_code_modifier.dart';
 import '../code_theme/code_theme.dart';
 import '../code_theme/code_theme_data.dart';
 import '../named_sections/parsers/abstract.dart';
+import '../symbols.dart';
 import '../wip/autocomplete/popup_controller.dart';
 import 'editor_params.dart';
 import 'span_builder.dart';
@@ -260,7 +261,7 @@ class CodeController extends TextEditingController {
   String get fullText => _lastCode.text;
 
   set fullText(String fullText) {
-    _updateLastCodeIfChanged(_replaceTabsIfNeeded(fullText));
+    _updateLastCodeIfChanged(_replaceTabsWithSpacesIfNeeded(fullText));
     super.value = TextEditingValue(text: _lastCode.visibleText);
   }
 
@@ -378,9 +379,9 @@ class CodeController extends TextEditingController {
     );
   }
 
-  String _replaceTabsIfNeeded(String text) {
+  String _replaceTabsWithSpacesIfNeeded(String text) {
     if (modifiers.contains(const TabModifier())) {
-      return text.replaceAll('\t', '  ');
+      return text.replaceAll(Symbols.tab, Symbols.space * params.tabSpaces);
     }
     return text;
   }
--- a/lib/src/code_field/code_controller.dart
+++ b/lib/src/code_field/code_controller.dart
@@ -16,12 +16,13 @@ import '../code_modifiers/tab_code_modifier.dart';
 import '../code_theme/code_theme.dart';
 import '../code_theme/code_theme_data.dart';
 import '../named_sections/parsers/abstract.dart';
-import '../symbols.dart';
 import '../wip/autocomplete/popup_controller.dart';
 import 'editor_params.dart';
 import 'span_builder.dart';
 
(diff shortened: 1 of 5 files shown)


test/src/code_field/code_controller_tab_test.dart
```diff
@@ -0,0 +1,40 @@
...
+
+    test(
+      'not applied if TabModifier is not present',
+      () {
+        final controller = CodeController(
+          text: textWithTabs,
+          language: go,
+          modifiers: [],
+        );
+        expect(controller.fullText.contains(Symbols.tab), true);
```
alexeyinkin, line 36: Its better to test this for the full match with the source.

test/src/code_field/code_controller_tab_test.dart
```diff
@@ -0,0 +1,40 @@
...
+  group('Tab replacement', () {
+    test(
+      'applied if TabModifier is present',
+      () {
+        final controller = CodeController(
+          text: textWithTabs,
+          language: go,
+          modifiers: [const TabModifier()],
+        );
+        expect(controller.fullText.contains(Symbols.tab), false);
```
alexeyinkin, line 24: It's better to test for the full match with the string with spaces.
Also test with different number of spaces per tab.

lib/src/code_field/code_controller.dart
```diff
@@ -378,6 +379,13 @@ class CodeController extends TextEditingController {
     );
   }
 
+  String _replaceTabsWithSpacesIfNeeded(String text) {
+    if (modifiers.contains(const TabModifier())) {
+      return text.replaceAll(Symbols.tab, Symbols.space * params.tabSpaces);
```
alexeyinkin, line 384: I think constants are an overkill here. Tab is clear from `\t`, and a space is clear from multiplying to `tabSpaces`.

alexeyinkin (changes requested):
Thanks.

lib/src/code_field/code_controller.dart
nausharipov: Constants seem more reliable to me. A space symbol ' ' can be accidentally edited like '  ' and we wouldn't notice the difference.

alexeyinkin: This is why IDEs have monospace fonts. This is why we have tests. And this is why we use `backticks` on GitHub.

Still I guess `\t` is explicit enouth.
And for space do you think we can use `charcode` package if you still want a constant?

`String.fromCharCode($space)`

We already use it elsewhere in our editor.

nausharipov: It's just my inexperience, but I didn't know before that '\t' is a tab. `fromCharCode` seems like a complication to me. 

1. I can move the tab and space constants from `Symbols.dart` into the code controller.
2. I can just comply and use '\t' and ' '.

alexeyinkin: Constants should not be created just to move all literals to the top. They are good to:
1. Reuse something that can fall out of sync (like size).
2. Extract something long to make its usage shorter (like a long string).
3. Allow to find all uses of something (like a URL).

Nothing of this applies here.

1. If it is a space, it will always be a space.
2. `' '` is shorter than any constant we can make of it.
3. Searching for "all spaces" is pointless.

Also `const _space = ' '; return _space;` is just as insecure as `return ' ';` but just is more complicated.

So I still prefer `return text.replaceAll('\t', ' ' * params.tabSpaces);`
Also `'\t'` for tab is a common knowledge.

My second best is:
`return text.replaceAll('\t', String.fromCharCode($space) * params.tabSpaces);`
It has the benefit that we know that it is a space because it is defined in a package. No one will fiddle with it replacing a space with an unbreakable space that still looks like `' '`.

My third best is this although I don't like it:
```dart
const _tab = '\t';
const _space = ' ';
```

nausharipov: I like the 3rd option after all. It is also easier to read: `replaceAll(_tab, _space * params.tabSpaces)` vs `replaceAll('\t', ' ' * params.tabSpaces)`.

alexeyinkin (approved):

PR #70, opened 2022-09-20
Issue45 python foldable block parser

### LINKED ISSUE #45: Parse collapsible blocks for Python

Malarg:
resolves #45


lib/src/folding/foldable_block_type.dart
```diff
@@ -8,6 +8,9 @@ enum FoldableBlockType {
   /// ( )
   parentheses,
 
+  //just for python
+  spaces,
```
alexeyinkin, line 12: Let's call this `indentation` because this can also originate from tabs (although we may indeed only support spaces for now).

lib/src/code/code.dart
```diff
@@ -3,9 +3,12 @@ import 'dart:math';
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:highlight/highlight_core.dart';
+import 'package:highlight/languages/python.dart';
 
 import '../folding/foldable_block.dart';
 import '../folding/parsers/highlight.dart';
+import '../folding/parsers/python_parser.dart';
+import '../folding/parsers/spaces_foldable_block_parser.dart';
```
alexeyinkin, line 11: The existing parsers' files are `abstract.dart` and `highlight.dart`.
We should either rename them to include `_parser` or `_foldable_block_parser`, or name this one `python.dart`.

Also the two new parsers' file names must be consistent with each other.

lib/src/folding/parsers/spaces_foldable_block_parser.dart
```diff
@@ -0,0 +1,109 @@
+import 'dart:collection';
+
+import 'package:charcode/ascii.dart';
+import '../foldable_block.dart';
+import '../foldable_block_type.dart';
+
+import 'abstract.dart';
+
+///A parser for foldable blocks from spaces count at start of each line.
+class SpacesFoldableBlockParser extends AbstractFoldableBlockParser {
```
alexeyinkin, line 10: Let's call this `IndentationFoldableBlockParser` because tabs are also a valid chars for this.
If you think you can add support for tabs quickly, then add it please. Otherwise please file an issue to support tabs for this parser.

lib/src/folding/parsers/spaces_foldable_block_parser.dart
```diff
@@ -0,0 +1,109 @@
+import 'dart:collection';
+
+import 'package:charcode/ascii.dart';
+import '../foldable_block.dart';
+import '../foldable_block_type.dart';
+
+import 'abstract.dart';
+
+///A parser for foldable blocks from spaces count at start of each line.
```
alexeyinkin, line 9: In English, it's "space count".
Also "indentation" is shorter than "space count at start of".

Add a space after `///` please.

lib/src/folding/parsers/spaces_foldable_block_parser.dart
```diff
@@ -0,0 +1,109 @@
+import 'dart:collection';
+
+import 'package:charcode/ascii.dart';
```
alexeyinkin, line 3: Package imports and relative imports should be separated by a blank line.
https://dart.dev/guides/language/effective-dart/style#ordering

lib/src/folding/parsers/spaces_foldable_block_parser.dart
```diff
@@ -0,0 +1,109 @@
...
+            type: FoldableBlockType.spaces,
+          ),
+        ),
+      );
+    });
+    blocks.addAll(resultBlocks);
+  }
+
+  List<int> _countWhitespacesInStartOfEachLine(String text) {
+    final lines = text.split('\n');
```
alexeyinkin, line 93: Splitting lines is expensive. We already have a list of lines from `Code` constructor. We should reuse it here.

lib/src/folding/parsers/spaces_foldable_block_parser.dart
```diff
@@ -0,0 +1,109 @@
...
+    final result = List.filled(lines.length, 0);
+    for (int i = 0; i < lines.length; i++) {
+      final line = lines[i];
+      var count = 0;
+      for (final char in line.runes) {
```
alexeyinkin, line 98: Looks like a great extension method on `String`.

alexeyinkin: Or even a getter like 'int get indentation`.

alexeyinkin: Oh, we already have `CodeLine` class. Then it should go there and not into an extension on `String`.

lib/src/folding/parsers/spaces_foldable_block_parser.dart
```diff
@@ -0,0 +1,109 @@
...
+///A parser for foldable blocks from spaces count at start of each line.
+class SpacesFoldableBlockParser extends AbstractFoldableBlockParser {
+  static const _kSeparatorLine = -1;
+
+  void parse(String text) {
+    _addWhitespacesBlocks(text);
+    finalize();
+  }
+
+  void _addWhitespacesBlocks(String text) {
```
alexeyinkin, line 18: This is too long a method and should be split.
You can accumulate open blocks in a field because this class is stateful anyway.
Good extraction candidates are:
`void _openBlock(int line, int indent)`
`void _closeBlocks(int line, int minIndent)`

lib/src/folding/parsers/spaces_foldable_block_parser.dart
```diff
@@ -0,0 +1,109 @@
...
+      }
+
+      if (nextExistingSpacesCount > currentSpacesCount) {
+        if (openBlocksBySpacesCount[currentSpacesCount] == null) {
+          openBlocksBySpacesCount[currentSpacesCount] =
+              List.empty(growable: true);
+        }
+
+        openBlocksBySpacesCount[currentSpacesCount]!.add(
+          FoldableBlock(
```
alexeyinkin, line 51: Do not reuse this class just because you need 1 of its 3 fields. One of the many reasons is that we may add fields there, and so this place has one more alien reason to change.

The parent solves this with `_StartedFoldableBlock` class.

lib/src/folding/parsers/spaces_foldable_block_parser.dart
```diff
@@ -0,0 +1,109 @@
...
+            );
+          }
+        });
+        openBlocksBySpacesCount.removeWhere(
+          (key, value) => key >= nextExistingSpacesCount,
+        );
+      }
+    }
+    openBlocksBySpacesCount.forEach((spacesCount, openedBlocks) {
+      resultBlocks.addAll(
```
alexeyinkin, line 79: If you are not going to filter these blocks afterwards, you can just add to the parent's `blocks`.

lib/src/code/code.dart
```diff
@@ -62,7 +65,12 @@ class Code {
     } else {
       final parser = HighlightFoldableBlockParser();
```
alexeyinkin, line 66: `Code` should not fiddle with the languages.
There should be a global factory that returns a parser for `Mode` (or even just a map, see `SingleLineComments` class). This factory should be called here.

This means that `AbstractFoldableBlockParser` should have an abstract `parse` method and define its arguments, so that all subclasses should follow that signature even if they do not use some of the arguments.

This also means that the Python parser should extend `AbstractFoldableBlockParser`.

alexeyinkin (changes requested):
So far I only reviewed the space parser, but enough changes are calling. Please make them, and then I will review the Python part. It would be inefficient now because those changes will affect it too.

alexeyinkin:
Let's stick to `indent` as a noun for all those identifiers.

`indent` as a `FoldableBlockType` constant.
`IndentFoldableBlockParser`.
`int get indent` in `CodeLine`.
`openBlocksByIndent`, `nextIndent`, etc. if you still need those variables.

test/src/folding/parsers/spaces_parser_test.dart
```diff
@@ -0,0 +1,127 @@
...
+            
+''',
+          expected: [
+            _FB(startLine: 2, endLine: 9, type: _T.spaces),
+            _FB(startLine: 3, endLine: 5, type: _T.spaces),
+            _FB(startLine: 7, endLine: 9, type: _T.spaces),
+            _FB(startLine: 8, endLine: 9, type: _T.spaces),
+          ],
+        ),
+      ];
```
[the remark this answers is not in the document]
Malarg: did not add test with multiline list, because it handles in python parser

lib/src/folding/parsers/spaces_foldable_block_parser.dart
Malarg: used just int value instead of creating new class

lib/src/code/code.dart
Malarg: awesome solution. Thanks!

test/src/folding/parsers/python_parser_test.dart
```diff
@@ -0,0 +1,207 @@
...
+              # comment
+              # another comment
+              # third comment
+                5,
+                6
+            ],
+            [
+                7,
+                8
+            ]
```
alexeyinkin, line 115: Please add another block with comments block indented further than the list content.
Please add a lone single line comment to make sure it does not form a block.

test/src/folding/parsers/python_parser_test.dart
```diff
@@ -0,0 +1,207 @@
...
+        ]''',
+          expected: [
+            _FB(startLine: 0, endLine: 14, type: FoldableBlockType.indent),
+            _FB(startLine: 1, endLine: 14, type: FoldableBlockType.indent),
+            _FB(startLine: 2, endLine: 14, type: FoldableBlockType.brackets),
+            _FB(startLine: 3, endLine: 9, type: FoldableBlockType.brackets),
+            _FB(
+              startLine: 4,
+              endLine: 6,
+              type: FoldableBlockType.singleLineComment,
```
alexeyinkin, line 125: We have `typedef _T` for such thing to fit in a single line.

test/src/folding/parsers/python_parser_test.dart
```diff
@@ -0,0 +1,207 @@
...
+          'Python. Pair characters in literals are ignored',
+          code: '''
+a = "{[(";
+b = ")]}";
+''',
+          expected: [],
+        ),
+        //
+        _Example(
+          'Python. Named comments',
```
alexeyinkin, line 141: Please expand this. Something like 'Service comments do not form blocks'

test/src/folding/parsers/python_parser_test.dart
```diff
@@ -0,0 +1,207 @@
...
+            serviceComments.map((e) => e.source).toSet(), codeLines);
+
+        expect(
+          pythonParser,
+          example.expected,
+          reason: '${example.name}, valid blocks',
+        );
+      }
+    });
+  });
```
alexeyinkin, line 188: When composing examples, please add diversity:
- Add comments after some lines that have `:` as their last significant character.
- Use trailing commas on some but not all multiline lists.
- Add global nodes other than `class`: variables, functions, etc.
- Use one-liners instead of some blocks like `if True: continue` to make sure they do not form blocks.

lib/src/folding/parsers/indent.dart
```diff
@@ -0,0 +1,96 @@
...
+import '../../code/code_line.dart';
+import '../foldable_block.dart';
+import '../foldable_block_type.dart';
+import 'abstract.dart';
+
+/// A parser for foldable blocks from lines indentation
+class IndentFoldableBlockParser extends AbstractFoldableBlockParser {
+  static const _kSeparatorLine = -1;
+
+  final openBlocksLinesByIndent = HashMap<int, List<int>>();
```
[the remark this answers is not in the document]
alexeyinkin: Can this map be private?

alexeyinkin (changes requested):
Thanks. Again I did not read the Python parser thoroughly. I will get back to it when the new questions are addressed.

lib/src/folding/parsers/indent.dart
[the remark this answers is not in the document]
Malarg: Looks like you are right. We really dont need list here. Thanks

lib/src/code/code.dart
```diff
@@ -57,11 +58,23 @@ class Code {
...
+      highlighted: highlighted,
+      commentsByLines: commentParser.getCommentsByLines(),
+    );
+
     if (highlighted == null) {
       foldableBlocks = const [];
     } else {
-      final parser = HighlightFoldableBlockParser();
-      parser.parse(highlighted, serviceCommentsNodesSet);
+      final parser = FoldableBlockParserFactory.provideParser(language ?? java);
```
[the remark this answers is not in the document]
Malarg: Unfortunately If we don't know about language at this place, I think it will be good idea to provide parser, which covers most of language we supports in playground (Java, Dart, Scala. May be Go, not sure). Of course this package can be used in another app, so exact we can do is make `mode` parameter at `Code` non nullable& I think it will be more clear solution

lib/src/folding/parsers/indent.dart
```diff
@@ -0,0 +1,96 @@
...
+          startLine: lineIndex,
+          endLine: endLine,
+          type: FoldableBlockType.indent,
+        ),
+      ),
+    );
+  }
+
+  List<int> _calculateLineIndents(List<CodeLine> lines) {
+    final result = List.filled(lines.length, 0);
```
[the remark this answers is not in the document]
Malarg: Agree with you, but in this constructor growable is false by default
```
external factory List.filled(int length, E fill, {bool growable = false});
```

lib/src/folding/parsers/indent.dart
```diff
@@ -0,0 +1,96 @@
...
+    List<CodeLine> lines,
+  ) {
+    _addWhitespacesBlocks(lines);
+    finalize();
+  }
+
+  void _addWhitespacesBlocks(List<CodeLine> lines) {
+    final lineIndents = _calculateLineIndents(lines);
+
+    int lastValuableLine = lineIndents.length - 1;
```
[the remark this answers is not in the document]
Malarg: renamed to lastNonEmptyLine. This variable helps parser to recognize case, where there are several separator lines at the end of code. It will reassign in the loop in this case.

lib/src/code/code.dart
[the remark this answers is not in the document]
alexeyinkin: Let's not do any parsing and folding at all without the language info. No folding is better than a wrong folding.

test/src/folding/parsers/spaces_parser_test.dart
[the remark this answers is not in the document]
alexeyinkin: 3. There is none added. Just pick any of the existing easy examples and use one space for indent in it.

4. There's still only an invalid test with insonsistent number of spaces. Please add a valid one:
```
level0
  level1
level0
 level1
```

6. Yes, this parser is not expected to handle this case correctly and produce "0-1" block for this. This is the point of the test -- to prove it is consistent in this incorrect behavior.

lib/src/folding/parsers/indent.dart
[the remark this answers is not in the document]
alexeyinkin: I'll keep this open for now and maybe deal with this later. If we are keeping this algo for now, this will also do.

lib/src/folding/parsers/indent.dart
```diff
@@ -9,88 +7,92 @@ import 'abstract.dart';
...
 
-      if (nextExistingIndent == _kSeparatorLine) {
-        lastValuableLine = i;
+      if (nextNonEmptyIndent == null) {
+        lastNonEmptyLine = i;
         break;
-      }
-
-      if (nextExistingIndent > currentLineIndent) {
+      } else if (nextNonEmptyIndent > currentLineIndent) {
```
alexeyinkin, line 42: No need for `else` after `break`.

lib/src/folding/parsers/indent.dart
```diff
@@ -0,0 +1,96 @@
...
+        lastValuableLine = i;
+        break;
+      }
+
+      if (nextExistingIndent > currentLineIndent) {
+        _openBlock(currentLineIndent, i);
+      }
+
+      if (nextExistingIndent < currentLineIndent) {
+        openBlocksLinesByIndent.forEach((spacesCount, openedBlocks) {
```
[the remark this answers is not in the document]
alexeyinkin: This is worse because it is puzzling.

To simplify this, extract the `for` loop into `_closeBlocks` method.

lib/src/folding/parsers/indent.dart
```diff
@@ -9,88 +7,92 @@ import 'abstract.dart';
...
-      if (nextExistingIndent < currentLineIndent) {
-        openBlocksLinesByIndent.forEach((spacesCount, openedBlocks) {
-          if (spacesCount >= nextExistingIndent) {
-            _closeBlocks(openedBlocks, i);
-          }
-        });
-        openBlocksLinesByIndent.removeWhere(
-          (key, value) => key >= nextExistingIndent,
+      } else {
+        _openBlocksLinesByIndent.forEachInvertedWhile((spacesCount, startLine) {
```
alexeyinkin, line 45: Please extract to a new `_closeBlocks` method.

test/src/folding/parsers/python_parser_test.dart
```diff
@@ -1,155 +1,196 @@
...
+            _FB(startLine: 0, endLine: 2, type: _T.indent),
+            _FB(startLine: 4, endLine: 6, type: _T.indent),
+          ],
+        ),
+
+        _Example(
+          'Python. Return only highlight blocks if it does not contain indents',
+          code: '''
+numbers = [1, 
+    2,
```
alexeyinkin, line 185: This is still an indent.

alexeyinkin (changes requested):
Thanks.

test/src/folding/parsers/python_parser_test.dart
```diff
@@ -0,0 +1,251 @@
...
+
+        expect(
+          pythonParser,
+          example.expected,
+          reason: '${example.name}, valid blocks',
+        );
+      }
+    });
+  });
+}
```
alexeyinkin, line 233: Please also add tests for:
- Imports. Try all forms of import: `import a`, `from a.b import c`, `from a import b as c`, etc. See if there are more. Check if Python supports breaking lines in an import statement and test that if it does.
- Single line comment block at the beginning of the file. This is especially important as all Apache files have a license comment header.

lib/src/folding/parsers/python.dart
```diff
@@ -0,0 +1,120 @@
...
+    );
+    final indentBlocks = getBlocksFromParser(
+      IndentFoldableBlockParser(),
+      highlighted,
+      serviceCommentsSources,
+      lines,
+    );
+
+    if (indentBlocks.isEmpty) {
+      return highlightBlocks;
```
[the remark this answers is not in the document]
Malarg: Oh. Understood that I wrote test incorrect. I already had test named 'Python. Return only highlight blocks if it does not contain indents', but it contained indents. Removed them

alexeyinkin:
Next time, for git to recognize a moved file, please make a separate commit that contains only the file renaming and no other changes. I am not sure this will work either, but the chance is higher.

image

lib/src/folding/parsers/python.dart
```diff
@@ -96,48 +49,88 @@ class PythonFoldableBlockParser extends AbstractFoldableBlockParser {
...
+  /// with priority to highlight blocks if blocks intersect.
+  ///
+  /// Lets consider this python code for example:
+  /// a = [     # 0
+  ///     1,    # 1
+  ///     2,    # 2
+  ///     3,    # 3
+  ///     4,]   # 4
+  /// Highlight block will return [0, 4].
+  /// Indent block will return [1, 3] block.
```
alexeyinkin, line 62: `[1, 4]`?

lib/src/folding/parsers/indent.dart
```diff
@@ -5,60 +5,98 @@ import '../foldable_block.dart';
...
+    _parse(lines);
     finalize();
   }
 
-  void _addWhitespacesBlocks(List<CodeLine> lines) {
-    final lineIndents = _calculateLineIndents(lines);
+  @override
+  void finalize() {
+    super.finalize();
+    _openBlocksLinesByIndent = {};
```
alexeyinkin, line 26: Can we do this in the end of `_parse()`? This will leave absolutely no chance of misuse while here there's still a chance a client will not call `finalize()`.

lib/src/folding/parsers/indent.dart
```diff
@@ -5,60 +5,98 @@ import '../foldable_block.dart';
 import '../foldable_block_type.dart';
 import 'abstract.dart';
 
-/// A parser for foldable blocks from lines indentation
+/// A parser for foldable blocks from lines indentation.
 class IndentFoldableBlockParser extends AbstractFoldableBlockParser {
-  final _openBlocksLinesByIndent = <int, int>{};
+  Map<int, int> _openBlocksLinesByIndent = <int, int>{};
+  List<int?> _linesIndents = [];
```
alexeyinkin, line 11: If you call `.clear()` on them, you can keep them `final`.

lib/src/folding/parsers/indent.dart
```diff
@@ -5,60 +5,98 @@ import '../foldable_block.dart';
...
+  void finalize() {
+    super.finalize();
+    _openBlocksLinesByIndent = {};
+    _linesIndents = [];
+  }
+
+  void _parse(List<CodeLine> lines) {
+    _linesIndents = _calculateLinesIndents(lines);
+    final significantIndentIndexes =
+        _SignificantIndentIndexes.fromLineIndents(_linesIndents);
```
alexeyinkin, line 33: 1. Make this a nullable static method.
2. Return null if no significant lines.
3. Make all fields non-nullable.
4. Remove all unnecessary `!` on them.

lib/src/folding/parsers/python.dart
```diff
@@ -96,48 +49,88 @@ class PythonFoldableBlockParser extends AbstractFoldableBlockParser {
...
+  /// We need to skip [1, 3] block because it is inside [0, 4] block.
+  List<FoldableBlock> _combineBlocks(
     List<FoldableBlock> highlightBlocks,
```
alexeyinkin, line 65: These arguments are better be named because the call looks like the order of the two does not matter but it matters a lot.

alexeyinkin (changes requested):
Thanks.

lib/src/folding/parsers/python.dart
Malarg: [0, 3] :) 
Thanks

lib/src/folding/parsers/indent.dart
Malarg: I cant because of this error: `Unsupported operation: Cannot clear a fixed-length list`

alexeyinkin: :(

alexeyinkin (approved):

PR #75, opened 2022-09-30
issue 47. tests for scala foldable blocks

Malarg:


test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -10,142 +10,126 @@ import 'package:flutter_code_editor/src/single_line_comments/single_line_comment
...
-
-        _Example(
-          'Java. Multiline pair-character blocks',
-          code: '''
-public class MyClass {
-  public static void main(
-    String[
-    ] args) {
-    }
+  _testJava();
```
alexeyinkin, line 16: This file is getting large. I suggest you split it by languages, make independent files with their `main`s, and put `parseAndCheck`, `FB`, and `T` to a shared file.

test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -10,142 +10,126 @@ import 'package:flutter_code_editor/src/single_line_comments/single_line_comment
...
+      const expectedBlocks = [
+        _FB(startLine: 0, endLine: 5, type: _T.braces),
+        _FB(startLine: 1, endLine: 3, type: _T.parentheses),
+        _FB(startLine: 2, endLine: 3, type: _T.brackets),
+        _FB(startLine: 3, endLine: 4, type: _T.braces),
+      ];
+      _parseAndCheck(mode: java, code: code, expected: expectedBlocks);
+    });
+
+    test('Java. Single-line pair-character blocks are ignored.', () {
```
alexeyinkin, line 43: No need to repeat the language here because 'Java' is already in the group name.
https://medium.com/flutter-senior/how-to-name-and-group-tests-in-a-dart-file-553e194bd8af

test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -158,119 +142,314 @@ import java.util.Date;
...
+    });
+
+    test('Scala. Imports at mid and the end, no newline in the end.', () {
+      const code = '''
+//
+package users
+import users._  // import everything from the users package
+class MyClass() {}
+import users.User  // import the class User
+import users.{User, UserPreferences}  // Only imports selected members''';
```
alexeyinkin, line 329: What happens if we break this line?
```scala
import users.{
  User,
  UserPreferences
}
```

test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -158,119 +142,314 @@ import java.util.Date;
...
+  def main(args: Array[String]): Unit =
+    println("Hello, Scala developer!")
+}''';
+      const expected = [
+        _FB(startLine: 0, endLine: 3, type: _T.braces),
+      ];
+      _parseAndCheck(mode: scala, code: code, expected: expected);
+    });
+
+    test('Scala. Multiline list form block.', () {
```
alexeyinkin, line 391: 'lists' or 'forms'.

test/src/folding/parsers/highlight_parser_test.dart
```diff
@@ -158,119 +142,314 @@ import java.util.Date;
...
+  true, // a boolean value
+  () => "an anonymous function returning a string"
+)''';
+      const expected = [
+        _FB(startLine: 0, endLine: 6, type: _T.parentheses),
+      ];
+      _parseAndCheck(mode: scala, code: code, expected: expected);
+    });
+
+    test('Scala. Match function form block.', () {
```
alexeyinkin, line 406: 'functions' or 'forms'.

alexeyinkin (changes requested):
Thanks.

test/src/folding/parsers/highlight_parser_test.dart
Malarg: Created an issue: https://github.com/akvelon/flutter-code-editor/issues/78

lib/src/folding/foldable_block.dart
```diff
@@ -23,10 +25,60 @@ class FoldableBlock with EquatableMixin {
...
   void sortByStartLine() {
     sort((a, b) => a.startLine - b.startLine);
   }
+
+  void joinIntersectedBlocks() {
+    if (length < 2) {
+      return;
+    }
+
+    _joinExistingIntersectedBlocks();
```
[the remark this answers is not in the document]
alexeyinkin: To write a test for this algo, imagine blocks like this:

```
|
|
|

|
| |
| | |
| | | |
| | | | |
| | | | |
| | | | |
| | |   |
| | |   |
| | |   |
| |     |
| |     |
|       |
|
```

test/src/folding/foldable_block_test.dart
```diff
@@ -0,0 +1,148 @@
+import 'package:flutter_code_editor/src/folding/foldable_block.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import 'parsers/test_executor.dart';
+
+void main() {
+  group('Foldable block test', () {
+    test('Join intersected blocks in empty list will not throw exception', () {
```
alexeyinkin, line 8: Please simplify test descriptions.
Like 'Empty -> Empty' in this case.

test/src/folding/foldable_block_test.dart
```diff
@@ -0,0 +1,148 @@
...
+      const blocks = <FoldableBlock>[];
+
+      final actual = blocks.deepCopy()..joinIntersecting();
+
+      expect(actual, blocks);
+    });
+
+    test('Join blocks in list with one block will not throw exception', () {
```
alexeyinkin, line 16: 'Single -> Single'

test/src/folding/foldable_block_test.dart
```diff
@@ -0,0 +1,148 @@
...
+
+      expect(actual, blocks);
+    });
+
+    test('Join intersecting blocks in list with two intersecting blocks', () {
+      const blocks = [
+        FoldableBlock(startLine: 0, endLine: 1, type: FBT.braces),
+        FoldableBlock(startLine: 1, endLine: 2, type: FBT.braces),
+      ];
+      const matcher = [
```
alexeyinkin, line 31: `Matcher` is a Dart class that does the comparison. This may be called 'expected' or alike.

test/src/folding/foldable_block_test.dart
```diff
@@ -0,0 +1,148 @@
...
+        FoldableBlock(startLine: 0, endLine: 2, type: FBT.union),
+      ];
+
+      final actual = blocks.deepCopy()..joinIntersecting();
+
+      expect(actual, matcher);
+    });
+
+    test('Join intersecting blocks in list with two not intersecting blocks',
```
alexeyinkin, line 40: Do we need this test if we have the one after it?

test/src/folding/foldable_block_test.dart
```diff
@@ -0,0 +1,148 @@
...
+      const matcher = [
+        FB(startLine: 0, endLine: 10, type: FBT.union),
+      ];
+
+      final actual = blocks.deepCopy()..joinIntersecting();
+
+      expect(actual, matcher);
+    });
+
+    test('Mixed blocks will join correctly', () {
```
alexeyinkin, line 79: This test mostly repeats the one above. You can add the last block from this test there, and the same inputs will be covered.

test/src/folding/foldable_block_test.dart
```diff
@@ -0,0 +1,148 @@
...
+        FB(startLine: 0, endLine: 5, type: FBT.braces),
+        FB(startLine: 1, endLine: 5, type: FBT.union),
+      ];
+
+      final actual = blocks.deepCopy()..joinIntersecting();
+
+      expect(actual, matcher);
+    });
+
+    test('Unsorted intersected blocks will not join', () {
```
alexeyinkin, line 111: I don't think we should test this case at all. We may keep the behavior on unsorted lists undefined. Then it will be easier to streamline the algorithm on sorted input.

Please add this expectation to the doc comment of the method.

test/src/folding/foldable_block_test.dart
```diff
@@ -0,0 +1,148 @@
...
+import 'package:flutter_test/flutter_test.dart';
+
+import 'parsers/test_executor.dart';
+
+void main() {
+  group('Foldable block test', () {
+    test('Join intersected blocks in empty list will not throw exception', () {
+      const blocks = <FoldableBlock>[];
+
+      final actual = blocks.deepCopy()..joinIntersecting();
```
alexeyinkin, line 11: Why deep copy? A shallow copy with `[...list]` will do.

lib/src/folding/foldable_block.dart
```diff
@@ -23,10 +25,44 @@ class FoldableBlock with EquatableMixin {
...
+
+  void joinIntersecting() {
+    if (length < 2) {
+      return;
+    }
+
+    for (int i = 1; i < length; i++) {
+      final currentBlock = this[i];
+      final previousBlock = this[i - 1];
+      if (currentBlock.intersects(previousBlock)) {
```
alexeyinkin, line 57: `intersects(other)` is not needed at all. Just check if `previous.endLine >= current.startLine && previous.endLine < current.endLine` or maybe make a method that encapsulates this (but I am struggling to come up with a name, maybe `intersectsWithEnd(int)`)

test/src/folding/foldable_block_test.dart
```diff
@@ -0,0 +1,148 @@
...
+        FB(startLine: 8, endLine: 10, type: FBT.imports),
+        FB(startLine: 1, endLine: 3, type: FBT.imports),
+      ];
+
+      final actual = blocks.deepCopy()..joinIntersecting();
+
+      expect(actual, blocks);
+    });
+
+    test('Containing children blocks will not join', () {
```
alexeyinkin, line 123: When testing for an undesired behavior, make it clear with a TODO comment.
```dart
    // TODO(username): Fix this
    test('Intersecting a non-immediate neighbor is not detected so far', () {
      const blocks = [
        FB(startLine: 0, endLine: 6, type: FBT.braces), //        0 intersects 3, not detected
        FB(startLine: 1, endLine: 3, type: FBT.parentheses), //   1
        FB(startLine: 4, endLine: 5, type: FBT.parentheses), //   2
        FB(startLine: 6, endLine: 12, type: FBT.braces), //       3
        FB(startLine: 7, endLine: 9, type: FBT.parentheses), //   4
        FB(startLine: 10, endLine: 11, type: FBT.parentheses), // 5
      ];

```

alexeyinkin: A diagram can also help right in a comment:

```
   0 1 2 3 4 5
 0 |
 1 | |
 2 | |
 3 | |
 4 |   |
 5 |   |
 6 |     |
 7       | |
 8       | |
 9       | |
10       |   |
11       |   |
12       |
```

alexeyinkin (changes requested):

test/src/folding/foldable_block_test.dart
Malarg: We do, because it's edge case. Two items are minimal available count for intersecting

test/src/folding/foldable_block_test.dart
Malarg: I'd like to test edge cases separately from another logic

lib/src/folding/foldable_block.dart
Malarg: This logic falls on this test:
```
        _Example(
          'Python. Nested highlight blocks',
          code: '''
a = [[
  5
  ]
]''',
          expected: [
            _FB(startLine: 0, endLine: 2, type: _T.brackets),
            _FB(startLine: 0, endLine: 3, type: _T.brackets),
          ],
        ),
```

alexeyinkin: Let's disallow two blocks to start on the same line because we can only have one folding button per line. Only the outer block of the two should survive.

test/src/folding/foldable_block_test.dart
alexeyinkin: Does it work differently internally? If no, then my experience is that less tests on similar inputs is preferred. This is because if we change the API we will have to refactor tests which is more tedious than refactoring code.

test/src/folding/foldable_block_test.dart
Malarg: Fixed tests

test/src/folding/foldable_block_test.dart
Malarg: Fixed tests

test/src/folding/foldable_block_test.dart
```diff
@@ -98,51 +74,48 @@ void main() {
...
+    // blocks[0] and blocks[3] should be joined, but they aren't.
     test('Containing children blocks will not join', () {
       const blocks = [
-        FB(startLine: 0, endLine: 6, type: FBT.braces),
-        FB(startLine: 1, endLine: 3, type: FBT.parentheses),
-        FB(startLine: 4, endLine: 5, type: FBT.parentheses),
-        FB(startLine: 6, endLine: 12, type: FBT.braces),
-        FB(startLine: 7, endLine: 9, type: FBT.parentheses),
-        FB(startLine: 10, endLine: 11, type: FBT.parentheses),
+        FB(startLine: 0, endLine: 6, type: FBT.braces), // 0
```
alexeyinkin, line 107: Align the digits please.

PR #87, opened 2022-10-18
issue69 implemented tabs to spaces replacing

Malarg:


lib/src/code_field/text_editing_value.dart
```diff
@@ -88,4 +90,91 @@ extension MyTextEditingValue on TextEditingValue {
...
+    );
+
+    return selection.isNormal ? result : result.reversed;
+  }
+
+  String get beforeSelection {
+    if (selection.baseOffset == -1 && selection.extentOffset == -1) {
+      return '';
+    }
+    final minSelection = min(selection.baseOffset, selection.extentOffset);
```
alexeyinkin, line 139: There's already `TextRange.start` that returns the min of the two.
Same is for `TextRange.end`.
Update here and below.

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
+import 'package:flutter/widgets.dart';
+import 'package:flutter_code_editor/src/code_field/text_editing_value.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+const _codeRowWithoutTabs = 'int a = 1;';
```
alexeyinkin, line 5: `_codeLineWithoutTabs`
Although I don't see a point in testing a single line separately.
For code without tabs, you can just reuse `_codeWithDoubleSpaces` and have one less `const`.

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+
+void main() {
+  group('Replace tabs with spaces', () {
+    test('Empty without selection', () {
+      TextEditingValue value = TextEditingValue.empty;
+      value = value.tabsToSpaces(_spaceCount);
+      
+      expect(value.text, '');
+      expect(value.selection.baseOffset, -1);
+      expect(value.selection.extentOffset, -1);
```
alexeyinkin, line 33: You can just compare to `TextEditingValue.empty`

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+    test('Empty with cursor at start', () {
+      TextEditingValue value = TextEditingValue.empty.copyWith(
+        selection: const TextSelection.collapsed(offset: 0),
+      );
+
+      value = value.tabsToSpaces(_spaceCount);
+      
+      expect(value.text, '');
+      expect(value.selection.baseOffset, 0);
+      expect(value.selection.extentOffset, 0);
```
alexeyinkin, line 45: You can compare the whole selection to `TextSelection.collapsed(offset: 0)`
Here and below.

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+    });
+
+    test('Code row without tabs without selection', () {
```
alexeyinkin, line 48: 'Code line...' here and below, if you still want to test a single line.

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      TextEditingValue value = const TextEditingValue(text: _codeRowWithoutTabs);
+      
+      value = value.tabsToSpaces(_spaceCount);
+      
+      expect(value.text, _codeRowWithoutTabs);
+      expect(value.selection.baseOffset, -1);
+      expect(value.selection.extentOffset, -1);
+    });
+
+    test('Code row without tabs with cursor at start', () {
```
alexeyinkin, line 58: 'the start'

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+  void method() {
+    print(field);
+  }
+}
+''';
+const _spaceCount = 2;
+
+
+void main() {
+  group('Replace tabs with spaces', () {
```
alexeyinkin, line 26: 1. You can further group 'Empty', 'Without tabs', and 'With tabs'
2. If combined test names do not read as a sentence, add periods after group names. https://medium.com/flutter-senior/how-to-name-and-group-tests-in-a-dart-file-553e194bd8af

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      );
+
+      value = value.tabsToSpaces(_spaceCount);
+
+      expect(value.text, _codeRowWithoutTabs);
+      expect(value.selection.baseOffset, cursorPosition);
+      expect(value.selection.extentOffset, cursorPosition);
+    });
+
+    test('Code row without tabs with selection', () {
```
alexeyinkin, line 99: 'Non-empty normalized selection'

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      );
+
+      value = value.tabsToSpaces(_spaceCount);
+
+      expect(value.text, _codeRowWithoutTabs);
+      expect(value.selection.baseOffset, selectionStart);
+      expect(value.selection.extentOffset, selectionEnd);
+    });
+
+    test('Code row without tabs with reversed selection', () {
```
alexeyinkin, line 117: 'Non-empty reversed selection'

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      
+      value = value.tabsToSpaces(_spaceCount);
+      
+      expect(value.text, _codeWithDoubleSpaces);
+      expect(value.selection.baseOffset, 0);
+      expect(value.selection.extentOffset, 0);
+    });
+
+    test('Code with tabs with cursor at middle', () {
+      const cursorPosition = _codeWithTabs.length ~/ 2;
```
alexeyinkin, line 159: Hardcode a number please so it can be easily verified.
Then there's no need for a comment below.

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      expect(value.text, _codeRowWithoutTabs);
+      expect(value.selection.baseOffset, 0);
+      expect(value.selection.extentOffset, 0);
+    });
+
+    test('Code row without tabs with cursor at middle', () {
+      const cursorPosition = 5;
+      TextEditingValue value = const TextEditingValue(
+        text: _codeRowWithoutTabs,
+        selection: TextSelection.collapsed(offset: cursorPosition),
```
alexeyinkin, line 75: Please test different `affinity` in some cases. Maybe here.

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -0,0 +1,223 @@
...
+      expect(value.selection.baseOffset, cursorPosition);
+      expect(value.selection.extentOffset, cursorPosition);
+    });
+
+    test('Code row without tabs with selection', () {
+      const selectionStart = 5;
+      const selectionEnd = 10;
+      TextEditingValue value = const TextEditingValue(
+        text: _codeRowWithoutTabs,
+        selection: TextSelection(
```
alexeyinkin, line 104: Please test `isDirectional` in some cases. Maybe here.

lib/src/code_field/text_editing_value.dart
```diff
@@ -88,4 +90,91 @@ extension MyTextEditingValue on TextEditingValue {
...
+    String afterSelection,
+  ) {
+    if (selection.baseOffset == -1 || selection.extentOffset == -1) {
+      return const TextSelection.collapsed(offset: -1);
+    }
+
+    final baseOffset = beforeSelection.length;
+    final extentOffset = baseOffset + inSelection.length;
+
+    final result = TextSelection(
```
alexeyinkin, line 125: Please use `TextSelection.copyWith` to preserve everything other than the offsets.

test/src/code_field/code_controller_tab_test.dart
```diff
@@ -20,6 +22,26 @@ public class MyClass {
...
+class MyClass {
+  void readOnlyMethod() {// [START section1]
+  }// [END section1]
+  // [START section2]
+  void method() {
+  }// [END section2]
+}
 ''';
 
   bool _areIdenticalTexts(String text1, String text2) {
```
alexeyinkin, line 47: Can we get rid of this and just use `expect(text1, text2)`?

test/src/code_field/code_controller_tab_test.dart
```diff
@@ -73,5 +95,70 @@ public class MyClass {
...
+        _areIdenticalTexts(
+          controller.text,
+          snippetWithDoubleSpaces +
+              ' ' * controller.params.tabSpaces * tabCount,
+        ),
+        true,
+      );
+    });
+
+    //TODO(Malarg): implement logic
```
alexeyinkin, line 122: Do these test fail now? If so, please do not include them in this branch.

alexeyinkin (changes requested):

test/src/code_field/code_controller_tab_test.dart
```diff
@@ -73,5 +95,70 @@ public class MyClass {
         );
       },
     );
+
+    test('works with several tab insertion', () {
+      final controller = CodeController(
+        text: snippetWithTabs,
+        language: java,
+        modifiers: [const TabModifier()],
```
alexeyinkin, line 103: Please also test that tabs are not replaced without `TabModifier`.

alexeyinkin, line 103: Please also make `TabModifier` do nothing because we no longer use its logic for replacement.

test/src/code_field/code_controller_tab_test.dart
Malarg: Im most cases except one :)

Malarg: Yes, deleted

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -21,203 +20,218 @@ class Foo {
...
+          ),
+        );
+      });
+
+      test('with cursor at the end', () {
+        const cursorPosition = _codeWithDoubleSpaces.length;
+        TextEditingValue value = TextEditingValue(
+          text: _codeWithDoubleSpaces,
+          selection: const TextSelection.collapsed(offset: cursorPosition)
+              .copyWith(isDirectional: true),
```
alexeyinkin, line 95: Per docs, a directional selection only makes sense for non-collapsed selection. That is why `collapsed` constructor does not have `isDirectional` argument.
Please test `isDirecional` in any of the non-collapsed selections below.

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -21,203 +20,218 @@ class Foo {
...
+          selection: TextSelection(
+            baseOffset: selectionStart,
+            extentOffset: selectionEnd,
+          ),
+        );
+
+        value = value.tabsToSpaces(_spaceCount);
+
+        expect(value.text, _codeWithDoubleSpaces);
+        expect(value.selection.baseOffset, selectionStart);
```
alexeyinkin, line 122: If you construct a selection here, you also implicitly test that `affinity` and `isDirectional` have not changed which is beneficial. This is why we should avoid testing individual properties of standard classes that can be compared as a whole.
Please check each test of individual properties if it can be changed to comparison to a `TextSelection` object.

alexeyinkin (changes requested):

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -21,148 +21,114 @@ class Foo {
...
+  void expectWillNotChanged(TextEditingValue value) {
+    var actual = value;
+    final expected = value.copyWith();
+
+    actual = actual.tabsToSpaces(_spaceCount);
+
+    expect(actual, expected);
+  }
+
   group('Replace tabs with spaces.', () {
```
alexeyinkin, line 33: Please rename to 'TextEditingValue.tabsToSpaces`.

test/src/code_field/tabs_to_spaces_test.dart
```diff
@@ -21,148 +21,114 @@ class Foo {
 const _spaceCount = 2;
 
 void main() {
+  void expectWillNotChanged(TextEditingValue value) {
```
alexeyinkin, line 24: 1. Please move this function inside the first group. Otherwise it is unclear what it actually does that should cause no change.
2. Please rename to `expectSame`

alexeyinkin (changes requested):

alexeyinkin (approved):

PR #95, opened 2022-10-26
Preserve selection when folding and unfolding blocks

### LINKED ISSUE #81: Preserve selection when folding and unfolding
Folding and unfolding code currently resets the cursor position and selection to none.

alexeyinkin:
- Resolves #81

### DIFF
--- a/lib/src/hidden_ranges/hidden_ranges.dart
+++ b/lib/src/hidden_ranges/hidden_ranges.dart
@@ -209,7 +209,7 @@ class HiddenRanges {
       }
     }
 
-    // upper == lower
+    // upper == lower - 1, or lower, or lower + 1
     final range = ranges[upperRange];
     switch (range.compareToPosition(position)) {
       case -1:
--- a/lib/src/hidden_ranges/hidden_ranges.dart
+++ b/lib/src/hidden_ranges/hidden_ranges.dart
@@ -178,7 +178,7 @@ class HiddenRanges {
 
     // Linear interpolation search.
     while (upperRange > lowerRange) {
-      // Full characters at the lower ange start and at the upper range end.
+      // Full characters at the lower range start and at the upper range end.
       final lowerChar = ranges[lowerRange].start;
       final upperChar = ranges[upperRange].end;
 
--- a/lib/src/hidden_ranges/hidden_ranges.dart
+++ b/lib/src/hidden_ranges/hidden_ranges.dart
@@ -160,6 +160,67 @@ class HiddenRanges {
     );
   }
 
+  /// Translates the [position] in the full text to the position in
+  /// the visible text.
+  ///
+  /// If [position] is hidden, returns the last visible position before it.
+  int cutPosition(int position) {
+    if (ranges.isEmpty || position < ranges.first.start) {
+      return position;
+    }
+
+    if (position >= ranges.last.end) {
+      return position - hiddenCharactersBeforeRanges.last;
+    }
+
+    int lowerRange = 0;
+    int upperRange = ranges.length - 1;
+
+    // Linear interpolation search.
+    while (upperRange > lowerRange) {
+      // Full characters at the lower ange start and at the upper range end.
+      final lowerChar = ranges[lowerRange].start;
+      final upperChar = ranges[upperRange].end;
+
+      final rangeIndex = lowerRange +
+          ((position - lowerChar) /
+                  (upperChar - lowerChar) *
+                  (upperRange - lowerRange))
+              .floor();
+
+      if (rangeIndex < lowerRange) {
+        return position - hiddenCharactersBeforeRanges[lowerRange];
+      }
+      if (rangeIndex > upperRange) {
+        return position - hiddenCharactersBeforeRanges[upperRange + 1];
+      }
+
+      final range = ranges[rangeIndex];
+
+      switch (range.compareToPosition(position)) {
+        case 0:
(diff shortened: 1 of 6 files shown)


lib/src/hidden_ranges/hidden_ranges.dart
```diff
@@ -160,6 +160,67 @@ class HiddenRanges {
...
+    if (position >= ranges.last.end) {
+      return position - hiddenCharactersBeforeRanges.last;
+    }
+
+    int lowerRange = 0;
+    int upperRange = ranges.length - 1;
+
+    // Linear interpolation search.
+    while (upperRange > lowerRange) {
+      // Full characters at the lower ange start and at the upper range end.
```
Malarg, line 181: typo "range"

lib/src/hidden_ranges/hidden_ranges.dart
```diff
@@ -160,6 +160,67 @@ class HiddenRanges {
...
+        return position - hiddenCharactersBeforeRanges[lowerRange];
+      }
+      if (rangeIndex > upperRange) {
+        return position - hiddenCharactersBeforeRanges[upperRange + 1];
+      }
+
+      final range = ranges[rangeIndex];
+
+      switch (range.compareToPosition(position)) {
+        case 0:
```
Malarg, line 201: MAy be will be more readable, if numbers will be changed on "equals", "upper", and "lower"

test/src/hidden_ranges/hidden_ranges_position_test.dart
```diff
@@ -4,28 +4,120 @@ import 'package:flutter_code_editor/src/hidden_ranges/hidden_range.dart';
...
 void main() {
+  group('HiddenRanges. cutPosition', () {
+    test('No ranges - No changes', () {
+      expect(
+        HiddenRanges.empty.cutPosition(7),
+        7,
+      );
+    });
+
+    test('Cut with visible beginning and end', () {
```
Malarg, line 20: This test does not cover case, where text contains hidden ranges at the start and at the end.

test/src/hidden_ranges/hidden_ranges_position_test.dart
```diff
@@ -4,28 +4,120 @@ import 'package:flutter_code_editor/src/hidden_ranges/hidden_range.dart';
...
+        6: 5,
+        7: 5,
+        8: 5,
+        9: 5,
+        10: 5,
+        11: 6,
+      };
+      int i = 0;
+
+      for (final example in examples.entries) {
```
Malarg, line 104: this loop repeat the same at 68 line. Suggest to extract

test/src/hidden_ranges/hidden_ranges_selection_test.dart
```diff
@@ -0,0 +1,211 @@
...
+        );
+        expect(
+          HiddenRanges.empty.recoverSelection(selection.reversed),
+          selection.reversed,
+        );
+      }
+    });
+
+    group('Collapsed', () {
+      test('before ranges', () {
```
Malarg, line 67: This 4 tests almost the same. Suggest to extract to a function and pass numbers as parameters

test/src/hidden_ranges/hidden_ranges_selection_test.dart
```diff
@@ -0,0 +1,211 @@
...
+        ),
+        TextSelection(
+          baseOffset: 5,
+          extentOffset: 10,
+          affinity: TextAffinity.upstream,
+          isDirectional: false,
+        ),
+      ];
+
+      for (final selection in selections) {
```
Malarg, line 45: May be this function helps to make this loop more clear
```
  void _expectTheSame(TextSelection selection) {
    expect(
      HiddenRanges.empty.cutSelection(selection),
      selection,
    );
    expect(
      HiddenRanges.empty.recoverSelection(selection),
      selection,
    );
  }
```

Malarg (changes requested):

test/src/hidden_ranges/hidden_ranges_selection_test.dart
alexeyinkin: This hides `expect` calls which is a bad pattern. They should be clearly visible. In 6 months when something breaks you will want to see `expect` on the top level.

test/src/hidden_ranges/hidden_ranges_position_test.dart
alexeyinkin: `+`

lib/src/hidden_ranges/hidden_ranges.dart
```diff
@@ -160,6 +160,67 @@ class HiddenRanges {
     );
   }
 
+  /// Translates the [position] in the full text to the position in
+  /// the visible text.
+  ///
+  /// If [position] is hidden, returns the last visible position before it.
+  int cutPosition(int position) {
```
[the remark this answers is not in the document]
alexeyinkin: This is longer. `cut...` and `recover...` fit well.

lib/src/hidden_ranges/hidden_ranges.dart
alexeyinkin: `-1`, `0`, and `1` are common for `sign`-like methods.

test/src/hidden_ranges/hidden_ranges_selection_test.dart
alexeyinkin: Merged with the test below.

PR #115, opened 2022-11-08
Change read-only section names (#110)

### LINKED ISSUE #110: Allow to change read-only sections after CodeController construction

alexeyinkin:
- Resolves #110

### DIFF
--- a/test/src/code_field/code_controller_readonly_sections_test.dart
+++ b/test/src/code_field/code_controller_readonly_sections_test.dart
@@ -1,0 +1,95 @@
+import 'package:flutter_code_editor/flutter_code_editor.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '../common/snippets.dart';
+
+void main() {
+  group('CodeController Read-only Sections', () {
+    test('No sections -> All editable', () {
+      final sectionGroups = <Set<String>>[
+        {},
+        {'nonexistent'},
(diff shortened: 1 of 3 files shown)


test/src/code_field/code_controller_readonly_sections_test.dart
```diff
@@ -0,0 +1,95 @@
...
+      );
+      controller.addListener(() {
+        notifiedCount++;
+      });
+
+      final readonly1 = controller.code.lines.lines.map((l) => l.isReadOnly);
+      expect(
+        readonly1,
+        [
+          false,
```
Malarg, line 47: Suppose we can make it more clear:
```
          false,  // class MyClass {
          false,  //   void method1() {
          true,   //     if (false) {// [START section1]
          true,   //       return;
          true,   //     }// [END section1]
          false,  //   }
          false,  //
          false,  //   void method2() {// [START section2]
          false,  //     return;
          false,  //   }// [END section2]
          false,  // }
          false,  //
```

Malarg (approved):

test/src/code_field/code_controller_readonly_sections_test.dart
alexeyinkin: That would introduce two locations that can fall out of sync and never be auto-detected.

PR #119, opened 2022-11-09
Issue103 pale readonly sections

alexeyinkin:


lib/src/code_field/span_builder.dart
```diff
@@ -38,10 +46,55 @@ class SpanBuilder {
...
+    if (_lineIndex >= code.lines.lines.length) {
+      _lineIndex = code.lines.lines.length - 1;
+    }
+  }
+
+  TextStyle? _applyReadonlyIfRequired(TextStyle? style) {
+    if (code.lines.lines[_lineIndex].isReadOnly) {
+      if (style == null) {
+        return textStyle?.paled();
+      } else {
```
alexeyinkin, line 75: No `else` after `return`.

lib/src/highlight/node.dart
```diff
@@ -51,4 +51,37 @@ extension MyNode on Node {
 
     return null;
   }
+
+  Node splitNewLines() {
+    children = children?.splitNewLines();
```
alexeyinkin, line 56: Are they mutable? 🤦

This method is misleading because it splits children but not this node. Also a node's text is currently split in the extension on `List<Node>` which is not right, it should be one in an extension on a node itself. So let's do this:

1. Make it split both this node and children. For this, return `List<Node>`.
2. If `value == null`, return a list of a single new `Node` with split `children`.
3. Otherwise return a list of new nodes split with newlines.

Neither way do we mutate a node.

lib/src/highlight/node.dart
```diff
@@ -51,4 +51,37 @@ extension MyNode on Node {
 
     return null;
   }
+
+  Node splitNewLines() {
```
alexeyinkin, line 55: `splitLines()`

lib/src/highlight/node.dart
```diff
@@ -51,4 +51,37 @@ extension MyNode on Node {
...
+  }
+}
+
+extension NodeList on List<Node> {
+  List<Node> splitNewLines() {
+    final result = <Node>[];
+
+    for (int i = 0; i < length; i++) {
+      final node = this[i];
+      final splittedValue = this[i].value?.split('\n');
```
alexeyinkin, line 67: The past tense of 'split' is just 'split'.

lib/src/highlight/node.dart
```diff
@@ -51,4 +51,37 @@ extension MyNode on Node {
...
+
+  Node splitNewLines() {
+    children = children?.splitNewLines();
+    return this;
+  }
+}
+
+extension NodeList on List<Node> {
+  List<Node> splitNewLines() {
+    final result = <Node>[];
```
alexeyinkin, line 63:
```dart
return map((node) => node.splitLines()).expand((nodes) => nodes).toList(growable: false);
```

lib/src/code_field/span_builder.dart
```diff
@@ -1,17 +1,25 @@
...
+  final Code code;
+  final CodeThemeData? theme;
+  final TextStyle? textStyle;
 
-  TextSpan build({
-    required Code code,
-    required CodeThemeData? theme,
-    TextStyle? textStyle,
-  }) {
+  int _lineIndex = 0;
```
alexeyinkin, line 13: `_visibleLineIndex`

lib/src/code_field/span_builder.dart
```diff
@@ -38,10 +46,55 @@ class SpanBuilder {
     required Node node,
     required CodeThemeData? theme,
   }) {
+    _updateLineIndex(node);
```
alexeyinkin, line 49: This should only be called on a node with `value` and not recursively.
And likely after this node is built.

lib/src/code_field/span_builder.dart
```diff
@@ -38,10 +46,55 @@ class SpanBuilder {
...
+  void _updateLineIndex(Node node) {
+    _lineIndex += node.getNewlineCount();
+
+    if (_lineIndex >= code.lines.lines.length) {
+      _lineIndex = code.lines.lines.length - 1;
+    }
+  }
+
+  TextStyle? _applyReadonlyIfRequired(TextStyle? style) {
+    if (code.lines.lines[_lineIndex].isReadOnly) {
```
alexeyinkin, line 72: This visible line index should be translated to a full line index.

lib/src/code_field/span_builder.dart
Malarg: Value already checks on null at `node.getNewlineCount` 
Why not recursively? Children also can contain value

lib/src/code_field/span_builder.dart
```diff
@@ -56,25 +56,26 @@ class SpanBuilder {
...
+    if (_visibleLineIndex >= code.lines.length) {
+      _visibleLineIndex = code.lines.length - 1;
     }
   }
 
-  TextStyle? _applyReadonlyIfRequired(TextStyle? style) {
-    if (code.lines.lines[_lineIndex].isReadOnly) {
+  TextStyle? _paleIfRequired(TextStyle? style) {
+    final revoverLineIndex =
+        code.hiddenLineRanges.revoverLineIndex(_visibleLineIndex);
```
alexeyinkin, line 73: `recoverLineIndex()`
And I usually name the variable `fullLineIndex`

lib/src/code_field/span_builder.dart
```diff
@@ -56,25 +56,26 @@ class SpanBuilder {
...
+    if (code.lines[revoverLineIndex].isReadOnly) {
       if (style == null) {
```
alexeyinkin, line 75: `return (style ?? textStyle).paled();`?

lib/src/hidden_ranges/hidden_line_ranges.dart
```diff
@@ -7,18 +7,81 @@ class HiddenLineRanges with EquatableMixin {
   final int fullLineCount;
   final int visibleLineCount;
 
-  const HiddenLineRanges({
+  final List<int> _recoveredLines;
+  final Map<int, int> _cutLines;
```
alexeyinkin, line 11: I guess a non-growable list is faster than a map.
```dart
final List<int?> _fullToVisible;
final List<int> _visibleToFull;
```

lib/src/hidden_ranges/hidden_line_ranges.dart
```diff
@@ -7,18 +7,81 @@ class HiddenLineRanges with EquatableMixin {
...
+
+  factory HiddenLineRanges({
+    required List<LineNumberingBreakpoint> breakpoints,
+    required int fullLineCount,
+    required int visibleLineCount,
+  }) {
+    final recoveredLines = <int>[];
+    final cutLines = <int, int>{};
+    for (var i = 0; i <= fullLineCount; i++) {
```
alexeyinkin, line 20: What is the time-complexity of this loop?

lib/src/hidden_ranges/hidden_line_ranges.dart
```diff
@@ -7,18 +7,81 @@ class HiddenLineRanges with EquatableMixin {
...
+  static const empty = HiddenLineRanges._(
     breakpoints: [],
     fullLineCount: 1,
     visibleLineCount: 1,
+    recoveredLines: [0],
+    cutLines: {},
   );
 
+  int? cutLineIndexIfVisible(int lineIndex) {
+    if (lineIndex < 0) {
```
alexeyinkin, line 54: I think it's OK to drop these checks and throw if out of bounds. We now have tests for out of range lines, but we did support that only because with breakpoints it was easier to allow those. We can drop that support if it is now easier to only support valid ranges.

lib/src/highlight/result.dart
```diff
@@ -41,8 +41,28 @@ extension MyResult on Result {
...
 
-  Result splitNewLines() {
-    nodes = nodes?.splitNewLines();
-    return this;
+  Result splitLines() {
+    return copyWith(
+      nodes: nodes
+          ?.map((n) => n.splitLines())
+          .expand((e) => e)
+          .toList(growable: false),
```
alexeyinkin, line 49: Duplicated code.
I liked the idea of an extension on `List<Node>` for that.

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -0,0 +1,67 @@
...
+          highlighted.splitLines().nodes!.any((node) => node.hasSeveralLines()),
+          false,
+        );
+      });
+    });
+  });
+}
+
+extension on Node {
+  bool hasSeveralLines() {
```
alexeyinkin, line 60: `getNewlineCount()` is recursive, so you don't need this.

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -0,0 +1,67 @@
...
+        Result? highlighted;
+
+        final mode = dart;
+        highlight.registerLanguage('language', mode);
+
+        highlighted = highlight.parse(text, language: 'language');
+
+        expect(
+          highlighted.splitLines().nodes!.any((node) => node.hasSeveralLines()),
+          false,
```
alexeyinkin, line 52: This test is insufficient. It can succeed even if `splitLines()` just deletes everything.
We must check that no extra splits are there and nothing is lost.

lib/src/highlight/node.dart
```diff
@@ -52,36 +52,41 @@ extension MyNode on Node {
     return null;
   }
 
-  Node splitNewLines() {
-    children = children?.splitNewLines();
-    return this;
-  }
-}
+  List<Node> splitLines() {
+    Node createNode([String? val]) => copyWith(
```
alexeyinkin, line 56: Vague semantics. Remind me to discuss.

lib/src/highlight/node.dart
```diff
@@ -53,30 +53,32 @@ extension MyNode on Node {
...
     }
-
     return result;
   }
 
+  String toStringRecursive() {
+    final result = {};
+
+    result['className'] = className;
+    result['value'] = '\'$value\'';
```
alexeyinkin, line 75: `value == null ? null : '\'$value\'';`

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -1,67 +1,150 @@
...
+  'Single line comment': _Example(
+    text: '''
 public class MyClass {
   public void main() { //comment
   }
 }''',
+    expected: 
+'''
+{value: 'public '}
+{className: class, children: ({value: ''}, {className: keyword, children: ({value: 'class'})}, {value: ' '}, {className: title, children: ({value: 'MyClass'})}, {value: ' '})}
```
alexeyinkin, line 20: It's hard to read this.
I wish we had something like this:

```
{value: 'public '}
{className: class}
  {value: ''}
  {className: keyword}
    {value: 'class'}
...
```

Also with this format `value: null` will not be an obstacle.

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -1,67 +1,150 @@
...
 
 void main() {
   group('Highlight split lines test.', () {
-    examples.forEach((name, text) {
+    examples.forEach((name, example) {
       test(name, () {
         Result? highlighted;
-
         final mode = dart;
         highlight.registerLanguage('language', mode);
```
alexeyinkin, line 127: No need to register the language in the loop.

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -1,67 +1,150 @@
...
+        highlighted = highlight.parse(example.text, language: 'language');
```
alexeyinkin, line 128: `final highlighted = ...`

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -1,67 +1,150 @@
...
+{value: '}'}''',
+  ),
 
-  'Readonly comment': '''
+  'Readonly comment': _Example(
+    text: '''
 public class MyClass {
   public void main() { // readonly
   }
 }''',
```
alexeyinkin, line 39: This is no different from the previous example.

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -1,67 +1,150 @@
...
+{value: '
+'}
+{value: '}'}''',
+  ),
 
-  'Invalid code (Folded)': '''
+  'Invalid code (Folded)': _Example(
+    text: '''
 class MyClass {
   void readOnlyMethod() {
```
alexeyinkin, line 98: Use neutral identifiers if they are not interpreted.

test/src/hidden_ranges/hidden_line_ranges_test.dart
```diff
@@ -31,13 +31,13 @@ void main() {
   group('HiddenLineRanges.', () {
     group('cutLineIndexIfVisible', () {
```
alexeyinkin, line 32: 'cutLineIndexIfVisible, recoverLineIndex'

test/src/hidden_ranges/hidden_line_ranges_test.dart
```diff
@@ -55,6 +55,18 @@ void main() {
...
+        expect(midBreakpointsRanges.recoverLineIndex(3), 5);
+        expect(midBreakpointsRanges.recoverLineIndex(4), 6);
+        expect(midBreakpointsRanges.recoverLineIndex(5), 9);
+        expect(midBreakpointsRanges.recoverLineIndex(6), 10);
+        expect(midBreakpointsRanges.recoverLineIndex(7), 11);
+        expect(midBreakpointsRanges.recoverLineIndex(11), 15);
+        expect(midBreakpointsRanges.recoverLineIndex(12), 100);
+        expect(midBreakpointsRanges.recoverLineIndex(13), 101);
       });
     });
```
alexeyinkin, line 71: Now that out-of-range causes errors, we need a test for empty text that ensures `0 <-> 0`.

lib/src/hidden_ranges/hidden_line_ranges.dart
```diff
@@ -12,44 +12,44 @@ class HiddenLineRanges with EquatableMixin {
...
   }) {
-    final fullToVisible = <int?>[];
-    final visibleToFull = <int>[];
+    final fullToVisible = List<int?>.filled(fullLineCount, null);
+    final visibleToFull = List<int>.filled(fullLineCount, 0);
 
     int n = 0;
 
+    var fullToVisibleIndex = 0;
+    var visibleToFullIndex = 0;
```
alexeyinkin, line 22: `fullIndex`, `visibleIndex`, drop `n`

lib/src/highlight/node.dart
```diff
@@ -55,28 +55,40 @@ extension MyNode on Node {
...
           value: splitValue[i] + (i == splitValue.length - 1 ? '' : '\n'),
+          children: i == 0 ? children?.splitLines() : null,
         ),
       );
     }
     return result;
   }
 
-  String toStringRecursive() {
+  String toStringRecursive([String accumulatedResult = '', int indent = 0]) {
```
alexeyinkin, line 72: Can you do this without `accumulatedResult`?

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -38,58 +15,73 @@ public class MyClass {
...
+  {className: title}
+    {value: 'MyClass', children: empty}
+  {value: ' ', children: empty}
+{value: '{\n', children: empty}
+{value: '  public ', children: empty}
+{className: keyword}
+  {value: 'void', children: empty}
+{value: ' main() { ', children: empty}
+{className: comment}
+  {value: '// readonly', children: empty}
```
alexeyinkin, line 33: 'readonly' is non-neutral, it grabs attention and suggests what it is not. Use something like 'comment' for meaningless comments.

test/src/code_field/split_highlight_lines_test.dart
```diff
@@ -38,58 +15,73 @@ public class MyClass {
...
+{value: '  }\n', children: empty}
+{value: '}', children: empty}
+''',
   ),
 
   'Named sections': _Example(
     text: '''
 class MyClass {
-  void readOnlyMethod() {// [START section1]
+  void someMethod() {// [START section1]
```
alexeyinkin, line 43: You may keep it, but I use `method1` and `method2` for such cases. It is perfectly symmetric so no method is special. And in this case it also mirrors 'section1' and 'section2'.

Another good naming is `startTrailingEndTrailing` and `startSeparateEndTrailing` to emphasize what we test here, then section names should mirror that as well.

A good code is the one that raises least 'why?' questions.

test/src/hidden_ranges/hidden_line_ranges_test.dart
```diff
@@ -3,6 +3,11 @@ import 'package:flutter_code_editor/src/hidden_ranges/line_numbering_breakpoint.
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
+  final emptyTextBreakpointRanges = HiddenLineRanges(
+    breakpoints: [],
+    fullLineCount: 1,
+  );
```
alexeyinkin, line 9: For some historical reason `Code.empty` has an empty list instead of a list of one empty line. I guess we also need a test that it produces `HiddenLineRanges` with `fullLineCount == 1` and not `0`.

lib/src/code/code_lines.dart
```diff
@@ -1,7 +1,6 @@
-import 'package:characters/characters.dart';
 import 'package:collection/collection.dart';
 import 'package:equatable/equatable.dart';
-import 'package:meta/meta.dart';
+import 'package:flutter/material.dart';
```
alexeyinkin, line 3: `TextRange` is from 'dart:ui'

lib/src/highlight/node.dart
```diff
@@ -69,26 +69,22 @@ extension MyNode on Node {
...
-    // ignore: parameter_assignments
-    if (children != null) {
-      for (final child in children!) {
-        // ignore: parameter_assignments
-        accumulatedResult =
-            child.toStringRecursive(accumulatedResult, indent + 1);
-      }
+  String toStringRecursive([int indent = 0]) {
+    final sb = StringBuffer();
+    sb.write('${' ' * 2 * indent}${toMapString()}\n');
```
alexeyinkin, line 74: `sb.writeln(' ' * 2 * indent + toMapString());`

Issue #120, opened 2022-11-15
Read-only formatting bugs


alexeyinkin:
1. On line 5, the `{` is grey but should not be.
2. The read-only text is grayed entirely instead of halving alpha.

![image](https://user-images.githubusercontent.com/44893228/201978711-ddf19047-b519-487f-a6c0-d3a55686fce5.png)

alexeyinkin:
https://github.com/akvelon/flutter-code-editor/pull/121

PR #121, opened 2022-11-16
Issue120 fix readonly bugs

Malarg:


Malarg:
resolves #120

lib/src/code_field/span_builder.dart
```diff
@@ -32,67 +33,58 @@ class SpanBuilder {
...
+        parentStyle: style ?? parentStyle,
       ),
-      style: _paleIfRequired(style),
+      style: paledStyle,
     );
   }
 
   void _updateLineIndex(Node node) {
-    _visibleLineIndex += node.getNewlineCount();
+    _visibleLineIndex += '\n'.allMatches(node.value ?? '').length;
```
alexeyinkin, line 75: `node.getValueNewlineCount();`

test/src/code_field/span_builder_test.dart
```diff
@@ -190,31 +244,40 @@ public class MyClass {
...
     });
   });
 }
 
 class _Example {
-  final String name;
   final String text;
   final Mode? mode;
   final TextSpan expected;
+  Set<String> readonlySectionNames;
```
alexeyinkin, line 275: final

lib/src/code_field/span_builder.dart
```diff
@@ -32,67 +33,58 @@ class SpanBuilder {
...
 
   TextSpan _buildNode({
     required Node node,
     required CodeThemeData? theme,
+    TextStyle? parentStyle,
   }) {
-    _updateLineIndex(node);
-
     final style = theme?.styles[node.className];
+    final paledStyle = _paleIfRequired(style, parentStyle);
```
alexeyinkin, line 59: `(style ?? parentStyle ?? textStyle)`

alexeyinkin (changes requested):

lib/src/code_field/span_builder.dart
```diff
@@ -53,10 +54,10 @@ class SpanBuilder {
   TextSpan _buildNode({
     required Node node,
     required CodeThemeData? theme,
-    TextStyle? parentStyle,
+    TextStyle? ancestorStyle,
   }) {
-    final style = theme?.styles[node.className];
-    final paledStyle = _paleIfRequired(style, parentStyle);
+    final classStyle = theme?.styles[node.className];
+    final paledStyle = _paleIfRequired(classStyle ?? ancestorStyle);
```
alexeyinkin, line 60: `classStyle ?? ancestorStyle` is duplicated below.
It's better to have `final style = theme?.styles[node.className] ?? ancestorStyle`

lib/src/code_field/span_builder.dart
```diff
@@ -65,25 +66,25 @@ class SpanBuilder {
       children: _buildList(
         nodes: node.children,
         theme: theme,
-        parentStyle: style ?? parentStyle,
+        ancestorStyle: classStyle ?? ancestorStyle,
       ),
       style: paledStyle,
```
alexeyinkin, line 71: `paledStyle` is not a good name because it can be not pale.

`style: _paleIfRequied(style),` ?

alexeyinkin (changes requested):

lib/src/code/text_style.dart
```diff
@@ -19,7 +19,7 @@ extension TextStyleExtension on TextStyle {
 
     return copyWith(
       color: Color.fromARGB(
-        clr.alpha ~/ 2,
+        127,
```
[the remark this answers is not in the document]
Malarg: Because we should avoid cases where we are pale already paled style

lib/src/code_field/span_builder.dart
Malarg: It isn't good idea to pale style after updating line index. Renamed variable

lib/src/code_field/span_builder.dart
Malarg: fixed

lib/src/code_field/span_builder.dart
```diff
@@ -1,39 +1,40 @@
 import 'package:flutter/widgets.dart';
 import 'package:highlight/highlight_core.dart';
-import '/src/code/text_style.dart';
 
-import '../../src/highlight/node.dart';
+import '/src/code/text_style.dart';
```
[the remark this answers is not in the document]
Malarg: fixed

lib/src/code/text_style.dart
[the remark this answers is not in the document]
alexeyinkin: This should not be a problem because `ancestorStyle` is never paled and retains the original opacity. It is only paled before the actual use in `TextStyle`. Have you actually seen this causing problems?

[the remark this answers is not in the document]
Malarg: You are right. Returned old logic

PR #122, opened 2022-11-19
Readme updates

AKosolapov:
Readme file updates

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,8 @@
-# flutter-code-editor
+# Flutter Code Editor
 
-A multi-platform code editor supporting
-syntax highlighting,
-code folding,
-autocompletion,
-and partial locking.
+Flutter Code Editor is a multi-platform code editor supporting syntax highlighting, code blocks folding, autocompletion, named sections, read-only code blocks, hiding text to create code snippets, themes and many other features.
+
+![Pub Version](https://img.shields.io/pub/v/flutter_code_editor)
 
 ![Basic Example](https://raw.githubusercontent.com/akvelon/flutter-code-editor/main/example/images/factorial.png)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,8 @@
-# flutter-code-editor
+# Flutter Code Editor
 
-A multi-platform code editor supporting
-syntax highlighting,
-code folding,
-autocompletion,
-and partial locking.
+Flutter Code Editor is a multi-platform code editor supporting syntax highlighting, code blocks folding, autocompletion, named sections, read-only code blocks, hiding text to create code snippets, themes and many other features.
```
alexeyinkin, line 3: Better not list named sections here because this is a service feature of no use by itself. It is only a mean to implement other features.

README.md
```diff
@@ -125,8 +118,13 @@ usable as one. However, code folding and other features have impact on built-in
 
 ## Named Sections
 
-Many features rely on so-called named sections in the code.
-You define named sections with service comments like this:
+Flutter Code Editor supports `named sections` in the source code. `Named sections` are defined in the souce code by adding a tag that Flutter Code Editor recognizes.
```
alexeyinkin, line 121: Backticks create monospace font and are traditionally used for identifiers and not the emphasis. *italic* is used for emphasis.

README.md
```diff
@@ -149,67 +147,61 @@ final controller = CodeController(
...
+In many languages, the editor recognizes sequential import lines and an optional package line as one foldable block. To fold such blocks:
 
 ```dart
 controller.foldImports();
 ```
 
 ### Named Sections
 
-You can fold all blocks except those overlapping with given named sections:
+Folding all blocks except the specific `named sections` is useful to create `code snippets` that help a user focus on the specific sections. The benefit of folding all blocks except the `named sections` is that all source code is still there, can be expanded, copied or downloaded by the user.
```
alexeyinkin, line 194: Downloading hints that the editor itself has this feature while it does not.

README.md
```diff
@@ -227,25 +219,26 @@ To get the currently folded blocks, read `controller.code.foldedBlocks`
 
 ## Hiding Text
 
-You can hide all the code except a given named section:
+Hiding all blocks except the specific `named sections` is useful to create `code snippets` that help a user focus on the specific code.
+
+Hiding text helps user achieve more focus than folding code blocks.
+
+The benefit of hiding all blocks except the `named sections` is that full text is preserved and available via `fullText` property in the Flutter Code Editor, can be copied or downloaded by the user.
```
alexeyinkin, line 226: Hidden text cannot be copied by Ctrl-A, Ctrl-C.

README.md
```diff
@@ -149,67 +147,61 @@ final controller = CodeController(
...
 ```
 
-This locks the given sections. Any non-existent section names in this set are ignored.
-To make the code editable again, pass an updated set.
+This locks the given sections from modifications in the Flutter Code Editor. Any non-existent section names in this set are ignored. To make the code editable again, pass an updated set to `controller.readOnlySectionNames`.
 
-When using this feature, `text` and `value` properties cannot be used to change the text
-programmatically because they have the same effect as the user input,
-and so locking affects them as well.
+When using this feature, `text` and `value` properties cannot be used to change the text programmatically because they have the same effect as the user input, and so locking affects them as well.
```
alexeyinkin, line 168: Single line breaks are ignored in markdown so I use them to keep lines short. If possible, please do not delete them. This also minimizes false-positive diffs.

AKosolapov:
@alexeyinkin I just got to be able to look at it now... Thank you for helping with this!

PR #124, opened 2022-11-28
issue123 changes from base repo

Malarg:


nausharipov:
Is there an issue this PR needs to be linked to?

lib/src/code_modifiers/indent_code_modifier.dart
```diff
@@ -5,10 +5,10 @@ import 'package:flutter/widgets.dart';
 import '../code_field/editor_params.dart';
 import 'code_modifier.dart';
 
-class IntendModifier extends CodeModifier {
+class IntentModifier extends CodeModifier {
   final bool handleBrackets;
 
-  const IntendModifier({
+  const IntentModifier({
```
nausharipov, line 11: The file name is `indent_code_modifier`, while the class is renamed to `IntentModifier`.

lib/src/code_field/code_controller.dart
```diff
@@ -131,14 +102,11 @@ class CodeController extends TextEditingController {
     this.stringMap,
     this.params = const EditorParams(),
     this.modifiers = const [
-      IntendModifier(),
+      IntentModifier(),
```
alexeyinkin, line 105: `IndentModifier`

lib/src/code_field/code_controller.dart
```diff
@@ -361,10 +311,6 @@ class CodeController extends TextEditingController {
     //   newValue = newValue.copyWith(text: _spacesToMiddleDots(newValue.text));
     // }
```
alexeyinkin, line 312: Delete this too.

alexeyinkin:
Did you run the full comparison like this?

image

It can reveal minor differences that can be ironed out like this blank line. Also this would prevent the `IntentModifier` typo.

Please rerun this way.

alexeyinkin:
image

alexeyinkin (changes requested):
I ran the full diff and saw no more to borrow except on the screens above. So you do not need to full-diff again.

lib/src/code_modifiers/indent_code_modifier.dart
Malarg: file named such way in a base repo. @alexeyinkin What do you think? Are we still support compatibility with the base repo?

lib/src/code_field/code_controller.dart
Malarg: fixed

lib/src/code_field/code_controller.dart
Malarg: done

nausharipov (approved):

alexeyinkin (approved):

PR #141, opened 2023-01-11
Issue48 fallback parser for java

### LINKED ISSUE #48: Fallback collapsible blocks parser for Java

yescorp:
- Resolves #48


yescorp:
Will support the following types of foldable blocks:
- imports `package ` `import `
- Single line comments `//`
- Multiline comments `/* */`
- Braces `{ }`
- Brackets `[ ]`
- Parentheses `( )`

Does Java have another types of foldable blocks?

lib/src/folding/parsers/fallback.dart
```diff
@@ -99,30 +113,94 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
+              }
             }
           }
 
-          if (!_foundSingleLineComment && !foundImport) {
+          if (!_foundSingleLineComment &&
+              !foundImport &&
+              !_isPossibleCommentSequence) {
             setFoundImportTerminator();
           }
```
yescorp, line 175: There was a problem that if the length of a comment sequence is more than 1 character, we won't mark it as a single line comment.
Because isFoundImportTerminator has more priority when submitting a line and after we found first character of the line that is not a full comment we would immediately  mark it as an import terminator

lib/src/folding/parsers/fallback.dart
```diff
@@ -99,30 +113,94 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
       }
 
       switch (code) {
         case $lf: // Newline
+          _isInDoubleQuoteLiteral = false;
+          _isInSingleQuoteLiteral = false;
```
yescorp, line 181: I think it makes sense to make unclosed quotes affect only until the end of line.

lib/src/folding/parsers/fallback.dart
```diff
@@ -13,21 +14,33 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
 
+  String? _startedMultilineCommentWith;
+
   /// If in a string literal the last char was a backslash.
   bool _wasBackslash = false;
 
-  bool _isInSingleQuoteLiteral = false;
   bool _isInDoubleQuoteLiteral = false;
-  bool _foundServiceSingleLineComment = false;
+  bool _isInMultilineComment = false;
```
alexeyinkin, line 29:
```suggestion
  bool get _isInMultilineComment => _startedMulilineCommentWith != null;
```

lib/src/folding/parsers/fallback.dart
```diff
@@ -13,21 +14,33 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
   bool _isLineStart = true;
+  bool _isPossibleCommentSequence = false;
+  bool _foundServiceSingleLineComment = false;
+  bool _shouldEndMultilineComment = false;
 
   FallbackFoldableBlockParser({
     required this.singleLineCommentSequences,
     required this.importPrefixes,
-  }) : _tailLength = singleLineCommentSequences.map((s) => s.length).max;
+    this.multilineCommentSequences,
```
alexeyinkin, line 39:
```suggestion
    this.multilineCommentSequences = [],
```
Then make it non-nullable

lib/src/folding/parsers/fallback.dart
```diff
@@ -196,4 +276,24 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
+    List<Tuple2<String, String>>? multilineCommentSequences,
+    List<String> singleLineCommentSequences,
+  ) =>
+      [
+        singleLineCommentSequences.map((s) => s.length).max,
+        multilineCommentSequences
+                ?.map((e) => max(e.item1.length, e.item2.length))
+                .max ??
+            0
+      ].max;
```
alexeyinkin, line 298:
```suggestion
      [
        ...singleLineCommentSequences.map((s) => s.length),
        ...multilineCommentSequences
                .map((e) => max(e.item1.length, e.item2.length))
      ].max;
```

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+  String[
+  ] args) {
+  }
+}''',
+        expected: [
+          _FB(firstLine: 0, lastLine: 5, type: _T.braces),
+          _FB(firstLine: 1, lastLine: 4, type: _T.union),
+        ],
+        invalidBlocks: [],
+      ),
```
alexeyinkin, line 75: Looks redundant given the Example 1.

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+          InvalidFoldableBlock(endLine: 1, type: _T.braces),
+          InvalidFoldableBlock(startLine: 1, type: _T.brackets),
+          InvalidFoldableBlock(startLine: 1, type: _T.parentheses),
+          InvalidFoldableBlock(endLine: 2, type: _T.brackets),
+          InvalidFoldableBlock(startLine: 2, type: _T.brackets),
+          InvalidFoldableBlock(startLine: 3, type: _T.braces),
+          InvalidFoldableBlock(startLine: 4, type: _T.braces),
+        ],
+      ),
+      _Example(
```
alexeyinkin, line 110: Add blank lines here and below please.

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+        'Pair characters in comments are ignored.',
+        code: '''
+public class MyClass {
+/// {[(                  1
+/*                       2
+  )]}                    3
+*/                       4
+/// {                    5
```
alexeyinkin, line 125: I have forgotten the added benefit of this line if it had any.

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+        ],
+      ),
+      _Example(
+        'Single line comments in a row',
+        code: '''
+// License                                            0
+// License                                            1
+public class MyClass {
+
+/// Method                                          4
```
alexeyinkin, line 139: Align the line number column please.

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+
+/// Method                                          6
```
alexeyinkin, line 141: Make neutral?

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+// comment
+}''',
+        expected: [
+          _FB(firstLine: 0, lastLine: 4, type: _T.braces),
+          _FB(firstLine: 1, lastLine: 2, type: _T.braces),
+        ],
+      ),
+
+      _Example(
+        'Service comment sequences do not form a foldable block.',
```
alexeyinkin, line 226:
```suggestion
        'A service comment sequence does not form a foldable block.',
```

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+      _Example(
+        'Pair symbols should be ignored inside a multiline comment',
+        code: '''
+/*[([       0
+])]*/       1
+''',
+        expected: [
+          _FB(firstLine: 0, lastLine: 1, type: _T.multilineComment),
+        ],
+      ),
```
alexeyinkin, line 245: Duplicate.

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+      _Example(
+        'Multiline comment after a brace',
+        code: '''
+class MyClass{            // 0
+
+} /*                         2
+* some weird comment         3
+*/                        // 4
```
alexeyinkin, line 253: Why they are not joined in a union?

lib/src/folding/parsers/fallback.dart
```diff
@@ -196,4 +276,24 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
 
   bool get _foundSingleLineComment =>
       _foundServiceSingleLineComment || foundSingleLineComment;
+
+  bool get isInMultilineComment =>
+      _isInMultilineComment && isMultilineCommentingEnabled;
+
+  bool get isMultilineCommentingEnabled =>
+      multilineCommentSequences != null &&
+      multilineCommentSequences!.isNotEmpty;
```
alexeyinkin, line 285: Private.

lib/src/folding/parsers/fallback.dart
```diff
@@ -196,4 +276,24 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
 
   bool get _foundSingleLineComment =>
       _foundServiceSingleLineComment || foundSingleLineComment;
+
+  bool get isInMultilineComment =>
+      _isInMultilineComment && isMultilineCommentingEnabled;
```
alexeyinkin, line 281: What if we just read `_isInMultilineComment` instead of this getter?

lib/src/folding/parsers/fallback.dart
alexeyinkin: Sounds good.

alexeyinkin (changes requested):

test/src/folding/parsers/java_parser_test.dart
yescorp: I was also confused.
But that's how Highlight parser does this.
I am trying to make Fallback parser's logic as close to Highlight parser as possible

test/src/folding/parsers/java_parser_test.dart
```diff
@@ -0,0 +1,390 @@
...
+        code: '''
+/*                                                  // 0
+ *                                                  // 1
+*/                                                  // 2
+
+package some.package;                               // 4
+
+import java.util.Arrays;                            // 6
+import org.apache.beam.sdk.Pipeline;                // 7
+import org.apache.beam.sdk.io.TextIO;               // 8
```
alexeyinkin, line 34: Make neutral.

lib/src/folding/parsers/fallback.dart
```diff
@@ -13,21 +14,33 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
   /// The size of a rolling window to remember processed characters.
   final int _tailLength;
 
+  String? _startedMultilineCommentWith;
+
   /// If in a string literal the last char was a backslash.
   bool _wasBackslash = false;
 
-  bool _isInSingleQuoteLiteral = false;
   bool _isInDoubleQuoteLiteral = false;
```
alexeyinkin, line 28: Preserve the two, then a blank line, then sort below.

lib/src/folding/parsers/fallback.dart
```diff
@@ -99,30 +113,94 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
                 }
                 break;
               }
+              if (c.contains(char)) {
+                _isPossibleCommentSequence = true;
+              }
+            }
+
+            if (isMultilineCommentingEnabled) {
```
alexeyinkin, line 121: No if.

lib/src/folding/parsers/fallback.dart
```diff
@@ -99,30 +113,94 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
                 }
                 break;
               }
+              if (c.contains(char)) {
+                _isPossibleCommentSequence = true;
```
alexeyinkin, line 117: Make local.

lib/src/folding/parsers/fallback.dart
```diff
@@ -99,30 +113,94 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
+                  if (!serviceCommentLines.contains(lineIndex)) {
+                    if (!foundImportTerminator) {
+                      startBlock(lineIndex, FoldableBlockType.multilineComment);
+                      _shouldEndMultilineComment = true;
+                    } else {
+                      // shouldn't start multiline comment block
+
+                      // class MyClass { /*
+                      // */
+                      // }
```
alexeyinkin, line 133: Check if this can be simplified if we make the highlight parser return union for 
```
{
} /*
*/
```

lib/src/folding/parsers/fallback.dart
```diff
@@ -99,30 +113,94 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
+                      _shouldEndMultilineComment = false;
+                    }
+                    _startedMultilineCommentWith = c.item1;
+                    _isInMultilineComment = true;
+                    break;
+                  }
+                }
+                if (c.item1.contains(char)) {
+                  _isPossibleCommentSequence = true;
+                }
```
alexeyinkin, line 143: Can we change the HighlightFoldableBlockParser so that /* */ comments are not treated as single line comments? Will this algorithm be simpler then?

lib/src/folding/parsers/fallback.dart
```diff
@@ -99,30 +113,94 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
+              }
+            }
+          }
+
+          if (isInMultilineComment) {
+            for (final c in multilineCommentSequences!) {
+              if (tail.endsWith(c.item2) &&
+                  _startedMultilineCommentWith == c.item1) {
+                if (!serviceCommentLines.contains(lineIndex)) {
```
alexeyinkin, line 152: Let us only allow `//`-like single line comments to be service comments.

lib/src/folding/parsers/fallback.dart
```diff
@@ -67,7 +80,8 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
     String tail = '';
 
     for (final code in text.runes) {
-      tail += String.fromCharCode(code); // ignore: use_string_buffers
+      final char = String.fromCharCode(code);
+      tail += char; // ignore: use_string_buffers
```
alexeyinkin, line 84: Maybe we can also accumulate future characters in `head` in a similar way so we can look 'into the future' to check if a character starts a comment?

This is complexity though. We should think carefully.

alexeyinkin (changes requested):

lib/src/folding/parsers/fallback.dart
yescorp: With `head` we can get rid of `tail` because they serve the same purpose.
It doesn't add a lot of complexity and allows us to get rid of a confusing flag.

I think it is better.

PR #143, opened 2023-01-16
Fix FoldableBlock.joinIntersecting(), fix newline handling after a service comment, export TextEditingValue.typed, add TextEditingValue.select

### LINKED ISSUE #136: CodeController.foldImports does not fold a multiline import in Go
Reproducible with MinimalWordCount example in Playground. In short:

```go
// license
// license
package main

import (
  "context"
  "fmt"
  "regexp"

  "github.com/apache/beam/sdks/v2/go/pkg/beam"
  "github.com/apache/beam/sdks/v2/go/pkg/beam/io/textio"
  "github.com/apache/beam/sdks/v2/go/pkg/beam/runners/direct"
  "github.com/apache/beam/sdks/v2/go/pkg/beam/transforms/stats"

  _ "github.com/apache/beam/sdks/v2/go/pkg/beam/io/filesystem/gcs"
  _ "github.com/apache/beam/sdks/v2/go/pkg/beam/io/filesystem/local"
)

```

alexeyinkin:
- Resolves #136

### DIFF
--- a/lib/src/code_field/text_editing_value.dart
+++ b/lib/src/code_field/text_editing_value.dart
@@ -152,4 +152,39 @@ extension TextEditingValueExtension on TextEditingValue {
     }
     return text.substring(selection.end);
   }
+
+  TextEditingValue typed(String text) {
+    final lengthDiff = text.length - selected.length;
+
+    return replaced(selection, text).copyWith(
+      selection: TextSelection.collapsed(offset: selection.end + lengthDiff),
+    );
+  }
+
+  TextEditingValue? select(Pattern pattern, [int start = 0]) {
+    if (pattern == '') {
+      throw AssertionError('Cannot search for an empty pattern');
+    }
+
+    final position = text.indexOf(pattern, start);
+
+    if (position == -1) {
+      return null;
+    }
+
+    final match = pattern.matchAsPrefix(text, position);
+
+    if (match == null) {
+      throw AssertionError('');
+    }
+
+    return TextEditingValue(
+      composing: composing,
+      selection: TextSelection(
+        baseOffset: position,
+        extentOffset: match.end,
+      ),
+      text: text,
+    );
+  }
 }
--- a/lib/src/folding/foldable_block.dart
+++ b/lib/src/folding/foldable_block.dart
@@ -111,7 +111,9 @@ extension FoldableBlockList on List<FoldableBlock> {
 
       // And fix every violation of the hierarchy by bubbling the block up,
       // removing non-ancestors from the working list, and joining when needed.
-      for (int ancestorIndex = ancestors.length - 1; --ancestorIndex >= 0;) {
+      for (int ancestorIndex = ancestors.length - 2;
+          ancestorIndex >= 0;
+          ancestorIndex--) {
         final ancestor = ancestors[ancestorIndex];
 
         if (ancestor.lastLine < bubble.firstLine) {
--- a/lib/src/folding/foldable_block.dart
+++ b/lib/src/folding/foldable_block.dart
@@ -55,6 +55,30 @@ class FoldableBlock extends InclusiveRange with EquatableMixin {
 
     return false;
   }
+
+  bool isSameLines(FoldableBlock other) {
+    return firstLine == other.firstLine && lastLine == other.lastLine;
+  }
+
+  FoldableBlock join(FoldableBlock other) {
+    return FoldableBlock(
+      firstLine: min(firstLine, other.firstLine),
+      lastLine: max(lastLine, other.lastLine),
+      type: _getJoinType(this, other),
+    );
+  }
+
+  static FoldableBlockType _getJoinType(FoldableBlock a, FoldableBlock b) {
+    if (a.type == FoldableBlockType.imports) {
+      return FoldableBlockType.imports;
+    }
+
+    if (b.type == FoldableBlockType.imports) {
+      return FoldableBlockType.imports;
+    }
+
+    return FoldableBlockType.union;
+  }
 }
 
 extension FoldableBlockList on List<FoldableBlock> {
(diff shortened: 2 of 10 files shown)


test/src/code_field/text_editing_value_test.dart
```diff
@@ -6,76 +6,76 @@ import '../common/lorem_ipsum.dart';
 
 void main() {
   group('TextEditingValue ', () {
-    test('wordAtCursorStart', () {
+    test('wordAtCursor', () {
```
alexeyinkin, line 9: Just changed the order of tests here, and it recognized them as changed lines.

lib/src/folding/foldable_block.dart
```diff
@@ -55,38 +55,102 @@ class FoldableBlock extends InclusiveRange with EquatableMixin {
...
+      // Throw the block into the current nesting hierarchy...
+      ancestors.add(bubble);
+      ancestorIndexToOverallIndex.add(overallIndex);
+
+      int bubbleIndexInAncestors = ancestors.length - 1;
+      int bubbleOverallIndex = overallIndex;
+
+      // And fix every violation of the hierarchy by bubbling the block up,
+      // removing non-ancestors from the working list, and joining when needed.
+      for (int ancestorIndex = ancestors.length - 1; --ancestorIndex >= 0;) {
```
yescorp, line 114: I think this way it is more readable.
```dart
for (int ancestorIndex = ancestors.length - 2; ancestorIndex >= 0; --ancestorIndex)
```
Also emphasizes on why the first iteration should not consider last element of the list.

lib/src/folding/foldable_block.dart
```diff
@@ -55,38 +55,102 @@ class FoldableBlock extends InclusiveRange with EquatableMixin {
 
     return false;
   }
+
+  bool isSameLines(FoldableBlock other) {
```
Malarg, line 59: `areSameLines`?

lib/src/folding/foldable_block.dart
```diff
@@ -55,38 +55,102 @@ class FoldableBlock extends InclusiveRange with EquatableMixin {
...
+    if (a.type == FoldableBlockType.imports) {
+      return FoldableBlockType.imports;
+    }
+
+    if (b.type == FoldableBlockType.imports) {
+      return FoldableBlockType.imports;
+    }
+
+    return FoldableBlockType.union;
+  }
```
Malarg, line 81:
```suggestion
  static FoldableBlockType _getJoinType(FoldableBlock a, FoldableBlock b) {
    final containsImports = a.isImports || b.isImports;
    return containsImports ? FoldableBlockType.imports : FoldableBlockType.union;
  }
```

lib/src/code_field/text_editing_value.dart
```diff
@@ -152,4 +152,39 @@ extension TextEditingValueExtension on TextEditingValue {
...
+    );
+  }
+
+  TextEditingValue? select(Pattern pattern, [int start = 0]) {
+    if (pattern == '') {
+      throw AssertionError('Cannot search for an empty pattern');
+    }
+
+    final position = text.indexOf(pattern, start);
+
```
nausharipov, line 170: Is there a rule for spacing? I expect no newline between related `position` initialization and and the `if` statement.

lib/src/folding/foldable_block.dart
alexeyinkin: Now imagine that we need to preserve the type if the blocks are of the same type. With this code, we would have to change lines, and also `?:` must be removed.

Those cut-offs are deliberate. They allow a lot of extension by just adding lines.

lib/src/folding/foldable_block.dart
alexeyinkin: `is` refers to the object and not its lines: `block.is...`

lib/src/folding/foldable_block.dart
alexeyinkin: `+`

lib/src/folding/foldable_block.dart
```diff
@@ -55,38 +55,104 @@ class FoldableBlock extends InclusiveRange with EquatableMixin {
...
-        );
-        removeAt(i);
-        i = max(i - 2, 0);
+    // A working list to lay down blocks to compare.
+    // As we iterate blocks, it will contain the current nesting hierarchy.
+    final ancestors = <FoldableBlock>[];
+    final ancestorIndexToOverallIndex = <int>[];
+
+    for (int overallIndex = 0; overallIndex < length; overallIndex++) {
+      FoldableBlock bubble = this[overallIndex];
```
nausharipov, line 103: A description for why it's named `bubble` might be useful.

lib/src/folding/foldable_block.dart
```diff
@@ -55,38 +55,104 @@ class FoldableBlock extends InclusiveRange with EquatableMixin {
...
   void sortByStartLine() {
     sort((a, b) => a.firstLine - b.firstLine);
   }
 
-  /// Joins intersecting blocks in list.
-  /// Expected that list is sorted by start line.
+  /// Joins intersecting blocks.
+  ///
+  /// This list must be sorted by firstLine.
   void joinIntersecting() {
```
nausharipov, line 92: Would helper methods improve readability?

test/src/code_field/text_editing_value_select_test.dart
```diff
@@ -0,0 +1,100 @@
...
+
+import 'package:flutter/widgets.dart';
+import 'package:flutter_code_editor/flutter_code_editor.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('TextEditingValue ', () {
+    test('select', () {
+      final examples = [
+        //
```
nausharipov, line 12: Is this empty comment deliberate?

lib/src/folding/foldable_block.dart
alexeyinkin: `+`

test/src/code_field/text_editing_value_select_test.dart
alexeyinkin: Yes, it preserves empty lines between examples.

PR #144, opened 2023-01-16
Issue48 fallback parser v2

### LINKED ISSUE #48: Fallback collapsible blocks parser for Java

yescorp:
- Resolves #48


lib/src/folding/parsers/fallback.dart
```diff
@@ -64,11 +71,11 @@ class FallbackFoldableBlockParser extends TextFoldableBlockParser {
...
     required CodeLines lines,
   }) {
-    String tail = '';
+    String line = '';
+    bool shouldEndMultilineComment = false;
 
     for (final code in text.runes) {
-      tail += String.fromCharCode(code); // ignore: use_string_buffers
-      tail = tail.substring(max(0, tail.length - _tailLength));
+      line += String.fromCharCode(code);
```
yescorp, line 78: I don't think using `StringBuffer` is going to be that beneficial here. Because we need to call `toString()` on it after each character anyway.
Will it have any impact?

PR #161, opened 2023-01-25
Can hide errors

alexrabin:
I have a use case in my project where I don't want to show any errors.


alexeyinkin:
- Done in #167

@alexrabin thanks for your effort. We did it in slightly another way.
In future please first open an issue to discuss changes.

PR #177, opened 2023-02-04
Foldable block matcher fix

### LINKED ISSUE #131: Foldable block matcher doesn't work properly
From the code below
```java
class MyClass {
  /// Calculates the factorial of the number.
  /// The number must be >= 0.
  static int factorial(int n) {
    if (n == 0) {
      return 1;
    }

    return n * factorial(n - 1);
  }

  public static void main(String[] args) {
    int num = 5;s
    System.out.println("Factorial of " + num + " is " + factorial(5));
  }
}
```
Steps to reproduce:

- Fold at the 4th line
- Move cursor to the end of 13th line
- Enter any character

`Expected`
- character is inserted to the text at cursor location.
- folded block preserve its folding

`Actual`:
- character is inserted at the end of 13th line
- folded block is unfolded
- cursor is moved to the end of 4th line

yescorp:
- Resolves #131

CC https://github.com/apache/beam/issues/25279

### DIFF
--- a/lib/src/gutter/gutter.dart
+++ b/lib/src/gutter/gutter.dart
@@ -127,6 +127,25 @@ class GutterWidget extends StatelessWidget {
             : () => codeController.foldAt(block.firstLine),
       );
     }
+
+    // Add folded blocks that are not considered as a valid foldable block,
+    // but should be folded because they were folded before becoming invalid.
+    for (final block in code.foldedBlocks) {
+      if (code.lines.length <= block.lastLine) {
+        continue;
+      }
+
+      final lineIndex = _lineIndexToTableRowIndex(block.firstLine);
+      if (lineIndex == null || lineIndex >= tableRows.length) {
+        continue;
+      }
+
+      tableRows[lineIndex].children![_foldingColumn] = FoldToggle(
+        color: style.textStyle?.color,
+        isFolded: true,
+        onTap: () => codeController.unfoldAt(block.firstLine),
+      );
+    }
   }
 
   int? _lineIndexToTableRowIndex(int line) {
--- a/lib/src/gutter/gutter.dart
+++ b/lib/src/gutter/gutter.dart
@@ -136,10 +136,6 @@ class GutterWidget extends StatelessWidget {
     // Add folded blocks that are not considered as a valid foldable block,
     // but should be folded because they were folded before becoming invalid.
     for (final block in code.foldedBlocks) {
-      if (code.lines.length <= block.lastLine) {
-        continue;
-      }
-
       final lineIndex = _lineIndexToTableRowIndex(block.firstLine);
       if (lineIndex == null || lineIndex >= tableRows.length) {
         continue;
--- a/test/src/code_field/code_controller_folding_editing_test.dart
+++ b/test/src/code_field/code_controller_folding_editing_test.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:highlight/languages/python.dart';
 
 import '../common/create_app.dart';
 import '../common/snippets.dart';
@@ -202,7 +203,7 @@ int n;
           });
 
           testWidgets('Braces. 3.', (wt) async {
-            const example = '\na{\n}';
+            const example = '\na{\n}\n';
             //               \ starting selection
             final controller = await pumpController(wt, example);
             controller.foldAt(1);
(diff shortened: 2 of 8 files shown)


test/src/code_field/code_controller_folding_editing_test.dart
```diff
@@ -146,9 +151,297 @@ int n;
...
+              language: python,
+            );
+
+            controller.foldAt(0);
+
+            expect(controller.value.text, 'a:\n\n');
+            //                                 \ selection
+            await wt.selectFromHome(3);
+
+            controller.value = controller.value.replacedSelection('\na');
```
alexeyinkin, line 275:
```suggestion
            controller.value = controller.value.replacedSelection('\n');
```
It a timer issue arises, debug it.

alexeyinkin (changes requested):

lib/src/gutter/gutter.dart
```diff
@@ -132,6 +132,25 @@ class GutterWidget extends StatelessWidget {
             : () => codeController.foldAt(block.firstLine),
       );
     }
+
+    // Add folded blocks that are not considered as a valid foldable block,
+    // but should be folded because they were folded before becoming invalid.
+    for (final block in code.foldedBlocks) {
+      if (code.lines.length <= block.lastLine) {
```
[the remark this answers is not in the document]
yescorp: This is an `early return condition` 
If we change the positions the condition should be the same I think.

test/src/code_field/code_controller_folding_editing_test.dart
```diff
@@ -203,6 +203,77 @@ int n;
             controller.value = controller.value.replacedSelection('');
             expect(controller.value, expected);
           });
+
+          testWidgets('Add new import after folded imports block', (wt) async {
+            const example = 'package example;\nimport java.util.date;\n';
```
alexeyinkin, line 208:
```suggestion
            const example = 'package example;\nimport java.util.Date;\n';
```

test/src/code_field/code_controller_folding_editing_test.dart
```diff
@@ -203,6 +203,77 @@ int n;
             controller.value = controller.value.replacedSelection('');
             expect(controller.value, expected);
           });
+
+          testWidgets('Add new import after folded imports block', (wt) async {
```
alexeyinkin, line 207:
```suggestion
          testWidgets('Adding an import after folded imports -> Unfold', (wt) async {
```

test/src/code_field/code_controller_folding_editing_test.dart
```diff
@@ -203,6 +203,77 @@ int n;
...
+            await wt.cursorEnd();
+            controller.value = controller.value.replacedSelection(inserted);
+            // TODO(yescorp): Fix selection, then fix tests
+            //  https://github.com/akvelon/flutter-code-editor/issues/182
+            expect(controller.value.text, example + inserted);
+            expect(controller.code.foldedBlocks.length, 0);
+          });
+
+          testWidgets(
+              'Add anything but imports line after folded imports block',
```
alexeyinkin, line 224:
```suggestion
              'Adding non-import after folded imports -> Does not unfold',
```

test/src/code_field/code_controller_folding_editing_test.dart
```diff
@@ -272,12 +343,13 @@ int n;
...
 
-            controller.value = controller.value.replacedSelection('\na');
+            controller.value = controller.value.replacedSelection('\n');
 
-            const expectedText = 'a:\n\na\n';
+            const expectedText = 'a:\n\n\n';
 
             expect(controller.value.text, expectedText);
             expect(controller.code.foldedBlocks.length, 1);
+            await wt.moveCursor(-1); // reset timer
```
alexeyinkin, line 352:
```suggestion
            await wt.moveCursor(-1); // Reset timer.
```

alexeyinkin (changes requested):

test/src/code_field/code_controller_folding_editing_test.dart
```diff
@@ -349,18 +350,14 @@ int n;
...
 
       group('Folded block is no longer recognizable', () {
-        group(
-            'Miltiline comment that started before folded block '
-            '1. Doesn\' unfold the block '
-            '2. The block is still unfoldable manually '
-            '3. The block is deletable', () {
-          testWidgets('Braces. 1.', (wt) async {
+        group('Miltiline comment that started before folded block', () {
+          testWidgets('1. Doesn\' unfold the block', (wt) async {
```
alexeyinkin, line 360:
```suggestion
          testWidgets('Does not unfold the block', (wt) async {
```

alexeyinkin (changes requested):

alexeyinkin (approved):

PR #187, opened 2023-02-15
History controller refactored

### LINKED ISSUE #132: Undo sets the cursor to the beginning of the document
https://user-images.githubusercontent.com/44893228/210773697-c867e1ed-88b4-4a3d-892d-1698a1526fcc.mp4

### LINKED ISSUE #186: Undo action issue
Undo action works incorrect: In this example after one undo action 2 actions happened: delete typed text and recover deleted
https://user-images.githubusercontent.com/17158960/218919532-b94ff00d-c74e-443e-9131-899dbd785764.mov

### LINKED ISSUE #193: For discussion: History controller interaction with folding
Seems like right now history controller snapshots the folding state of code as well.

https://user-images.githubusercontent.com/79095619/219439576-62ef5bc3-3acb-49f8-a08a-61ca9622c1ac.mov

IMO it is not practical at all. if the text inside folded block is not affected by it.

yescorp:
- Resolves #132 
- Resolves #186 
- Resolves #193

### DIFF
--- a/lib/src/history/code_history_controller.dart
+++ b/lib/src/history/code_history_controller.dart
@@ -83,8 +83,6 @@ class CodeHistoryController {
 
     lastCode = code;
     lastSelection = selection;
-
-    _removeLastRedundantSelectionOnlyChanges();
   }
 
   HistoryControllerAction _getAction({
@@ -92,21 +90,22 @@ class CodeHistoryController {
     required TextSelection selection,
     required bool isTextChanging,
   }) {
-    {
-      // If the line count is changed, we add record before and after edit.
-      if (code.lines.length != lastCode.lines.length) {
-        return HistoryControllerAction.newEntriesBeforeAndAfterEdit;
-      }
+    // If the line count is changed, we add record before and after edit.
+    if (code.lines.length != lastCode.lines.length) {
+      return HistoryControllerAction.newEntriesBeforeAndAfterEdit;
     }
 
-    {
-      // If change is caused by casual typing, we re-set the timer.
-      final isText1CharLonger = code.text.length == lastCode.text.length + 1;
-      final isTypingContinuous = isText1CharLonger &&
-          selection.hasMovedOneCharacterRight(lastSelection);
-      if (isTypingContinuous) {
-        return HistoryControllerAction.setTimer;
-      }
+    // If only selection changed, we add record before and after edit.
+    if (!isTextChanging && selection != lastSelection) {
+      return HistoryControllerAction.newEntriesBeforeAndAfterEdit;
+    }
+
+    // If change is caused by casual typing, we re-set the timer.
+    final isText1CharLonger = code.text.length == lastCode.text.length + 1;
+    final isTypingContinuous =
+        isText1CharLonger && selection.hasMovedOneCharacterRight(lastSelection);
+    if (isTypingContinuous) {
+      return HistoryControllerAction.setTimer;
     }
 
     // Any other change will create record after edit.
@@ -134,7 +133,21 @@ class CodeHistoryController {
 
         if (_isFullTextSame([last, lastMinus1, lastMinus2])) {
           stack.removeAt(stack.length - 2);
+          _currentRecordIndex--;
         }
+    }
+  }
+
+  void _removeLastIdenticalChanges() {
+    if (stack.length < 2) {
+      return;
+    }
+
+    final last = stack.last;
+    final preLast = stack[stack.length - 2];
+    if (last.code == preLast.code && last.selection == preLast.selection) {
+      stack.removeAt(stack.length - 1);
+      _currentRecordIndex--;
     }
   }
 
@@ -151,10 +164,7 @@ class CodeHistoryController {
   }
 
   void _dropRedoIfAny() {
-    final startIndexToRemove = _currentRecordIndex + 1;
-    if (startIndexToRemove < stack.length) {
-      stack.removeStartingAt(startIndexToRemove);
-    }
+    stack.removeStartingAt(_currentRecordIndex + 1);
   }
 
   void undo() {
@@ -185,23 +195,14 @@ class CodeHistoryController {
   }
 
   void _push() {
-    if (!lastSelection.isValid) {
-      // Do not create record for invalid selection
-      // as it is not considered to be standart user input.
-      return;
-    }
-
-    if (stack.isNotEmpty &&
-        stack.last.code.text == lastCode.text &&
-        stack.last.selection == lastSelection) {
-      // Do not create record if the last record is the same as the new one.
-      return;
-    }
     _dropRedoIfAny();
 
     _debounceTimer?.cancel();
     _pushRecord(_createRecord());
     _wasTextChanged = false;
+
+    _removeLastRedundantSelectionOnlyChanges();
+    _removeLastIdenticalChanges();
   }
 
   void _setTimer() {
--- a/lib/src/history/code_history_controller.dart
+++ b/lib/src/history/code_history_controller.dart
@@ -53,6 +53,9 @@ class CodeHistoryController {
   }) {
     if (isTextChanging) {
       _wasTextChanged = true;
+    }
+
+    if (isTextChanging || selection != lastSelection) {
       _dropRedoIfAny();
     }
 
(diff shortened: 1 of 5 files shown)


lib/src/history/code_history_controller.dart
```diff
@@ -46,45 +52,109 @@ class CodeHistoryController {
...
+
+    _removeLastRedundantSelectionOnlyChanges();
+  }
+
+  HistoryControllerAction _getAction({
+    required Code code,
+    required TextSelection selection,
+    required bool isTextChanging,
+  }) {
+    {
```
alexeyinkin, line 95: Delete redundant braces.

lib/src/history/code_history_controller.dart
```diff
@@ -46,45 +52,109 @@ class CodeHistoryController {
...
+      case 2:
+        // Drop the first one if only selection has changed.
+        if (_isFullTextSame([stack[0], stack[1]])) {
+          stack.removeAt(0);
+        }
+        break;
+
+      default:
+        // Check last, last-1, last-2. Drop last-1.
+        final last = stack.last;
```
alexeyinkin, line 131:
```suggestion
        final last = stack[stack.length - 1];
```
?

lib/src/history/code_history_controller.dart
```diff
@@ -46,45 +52,109 @@ class CodeHistoryController {
...
-    lastSelection = selection;
+    return true;
   }
 
   void _dropRedoIfAny() {
-    stack.removeStartingAt(_currentRecordIndex + 1);
+    final startIndexToRemove = _currentRecordIndex + 1;
+    if (startIndexToRemove < stack.length) {
+      stack.removeStartingAt(startIndexToRemove);
+    }
```
alexeyinkin, line 157: I cannot imagine an error that would require this. Maybe we need to decrement `_currentRecordIndex` in `_removeLastRedundantSelectionOnlyChanges`?
Otherwise please revert this and find the scenario when this fails, write a test, and then re-apply this.

lib/src/history/code_history_controller.dart
```diff
@@ -115,6 +185,20 @@ class CodeHistoryController {
   }
 
   void _push() {
+    if (!lastSelection.isValid) {
+      // Do not create record for invalid selection
+      // as it is not considered to be standart user input.
+      return;
+    }
```
alexeyinkin, line 192: Looks like a disguise for an erroneously losing selection when inserting a suggestion. Please try to preserve selection when inserting it.

lib/src/history/code_history_controller.dart
```diff
@@ -46,45 +52,109 @@ class CodeHistoryController {
...
+
+      default:
+        // Check last, last-1, last-2. Drop last-1.
+        final last = stack.last;
+        final lastMinus1 = stack[stack.length - 2];
+        final lastMinus2 = stack[stack.length - 3];
+
+        if (_isFullTextSame([last, lastMinus1, lastMinus2])) {
+          stack.removeAt(stack.length - 2);
         }
```
alexeyinkin, line 137: Also check if `lastMinus2 == lastMinus3`. This happens with multiple `enter` presses. Remove the check from `_push()`.

test/src/history/code_history_controller_test.dart
```diff
@@ -17,14 +17,60 @@ void main() {
...
         final controller = await pumpController(wt, MethodSnippet.full);
 
+        expect(controller.historyController.stack.length, 0);
+      });
+
+      testWidgets('Selection change creates 1 record after change', (wt) async {
+        final controller = await pumpController(wt, MethodSnippet.full);
+
+        // The stack is empty -> Selection change creates.
+        await wt.moveCursor(-1);
```
alexeyinkin, line 27: Write a comment: where is the cursor before this?

alexeyinkin (changes requested):

lib/src/history/code_history_controller.dart
yescorp: Yeah, decrementing fixed it.

lib/src/history/code_history_controller.dart
yescorp: Turns out, the method insertSelectedWord worked in a weird way.

test/src/history/code_history_controller_test.dart
```diff
@@ -252,73 +378,120 @@ void main() {
...
 
         await wt.sendRedo(); // No effect.
+        expect(controller.code.text, code4.text);
+        expect(controller.selection, selection4);
+        expect(controller.historyController.stack.last.code, code2);
+        expect(controller.historyController.stack.last.selection, selection2);
+      });
 
-        expect(controller.historyController.stack.length, 2);
+      testWidgets('Folded blocks match when undo/redo', (wt) async {
```
alexeyinkin, line 406: Start typing -> Fold -> Continue -> Undo -> Still folded

test/src/history/code_history_controller_test.dart
```diff
@@ -252,73 +378,120 @@ void main() {
...
-        expect(controller.historyController.stack.length, 2);
+        expect(controller.historyController.stack.length, 1);
       });
 
       testWidgets('Limit depth', (WidgetTester wt) async {
         final controller = await pumpController(wt, MethodSnippet.full);
         await wt.cursorEnd();
 
         // 1. Fill the limit.
-
```
alexeyinkin, line 316: Keep the blank line because the comment concerns not only the line below.

test/src/history/code_history_controller_test.dart
```diff
@@ -252,73 +378,120 @@ void main() {
...
-        for (int i = 0; i < CodeHistoryController.limit - 1; i++) {
+        for (int i = 0; i < CodeHistoryController.limit / 2 - 1; i++) {
           controller.value = controller.value.typed('a');
           await wt.moveCursor(-1); // Creates.
         }
```
alexeyinkin, line 486: Insert `ab` on each iteration so that it creates exactly 1 record. This way we avoid dependency on `limit` being even.

test/src/history/code_history_controller_test.dart
```diff
@@ -252,73 +378,120 @@ void main() {
...
         await wt.pumpAndSettle();
 
         controller.value = controller.value.copyWith(
           selection: const TextSelection.collapsed(offset: 0),
         );
-        expect(controller.historyController.stack.length, 2);
+        expect(controller.historyController.stack.length, 1);
       });
 
       testWidgets('Limit depth', (WidgetTester wt) async {
```
alexeyinkin, line 478: Simplify, only test that bottom records are deleted.

lib/src/history/code_history_controller.dart
```diff
@@ -46,41 +52,119 @@ class CodeHistoryController {
...
+        _setTimer();
+        break;
     }
 
-    if (!shouldSave) {
-      if (isTextChanging) {
-        _wasTextChanged = true;
-      }
+    lastCode = code;
+    lastSelection = selection;
```
alexeyinkin, line 88: Move to the timer's case.

alexeyinkin (changes requested):

test/src/history/code_history_controller_test.dart
```diff
@@ -286,102 +277,101 @@ void main() {
...
-        await wt.sendRedo();
-        expect(controller.historyController.stack.length, 6);
-        expect(controller.code.text, code1.text);
-        expect(controller.selection, selection1);
-
-        await wt.sendRedo();
-        expect(controller.historyController.stack.length, 6);
-        expect(controller.code.text, code2.text);
-        expect(controller.selection, selection2);
+        for (int i = manualHistoryRecords.length - 2; i >= 0; i--) {
```
alexeyinkin, line 334:
```suggestion
        for (int i = manualHistoryRecords.length - 1; --i >= 0; ) {
```

test/src/history/code_history_controller_test.dart
```diff
@@ -480,17 +469,22 @@ void main() {
...
-          controller.value = controller.value.typed('a');
-          await wt.moveCursor(-1); // Creates.
+
+        Code? firstCode;
+        TextSelection? firstSelection;
+
+        for (int i = 0; i < CodeHistoryController.limit - 2; i++) {
+          controller.value = controller.value.typed('ab'); // Creates.
+          firstCode ??= controller.code;
+          firstSelection ??= controller.selection;
```
alexeyinkin, line 479: codeAfterFirstEdit, selectionAfterFirstEdit
// Set after the first change.

alexeyinkin (changes requested):

alexeyinkin (approved):

PR #238, opened 2023-05-17
Issue225 fix suggestion error

Malarg:


lib/src/wip/autocomplete/popup_controller.dart
```diff
@@ -23,6 +23,10 @@ class PopupController extends ChangeNotifier {
 
   int get selectedIndex => _selectedIndex;
 
+  void reset() {
```
alexeyinkin, line 26: Add doc comment of why this is needed.
And a link to the bug in https://github.com/google/flutter.widgets

alexeyinkin (changes requested):

PR #243, opened 2023-06-05
issue 242 fixed bug: Search selection jumped when change search pattern that also fit to current selection

### LINKED ISSUE #242: Search selection jumps when change search pattern that also fit to current selection

Malarg:
- Resolves #242

issue 242 fixed bug: Search selection jumped when change search pattern that also fit to current selection

### DIFF
--- a/lib/src/code_field/search_result_highlighted_builder.dart
+++ b/lib/src/code_field/search_result_highlighted_builder.dart
@@ -4,13 +4,13 @@ import '../search/result.dart';
 import '../search/search_navigation_state.dart';
 
 @visibleForTesting
-const matchBackgroundColor = Colors.green;
+const matchBackgroundColor = Color.fromARGB(255, 255, 235, 0);
 
 @visibleForTesting
 const searchTextColor = Colors.black;
 
 @visibleForTesting
-const currentMatchBackgroundColor = Colors.yellow;
+const currentMatchBackgroundColor = Color.fromARGB(255, 255, 150, 50);
 
 class SearchResultHighlightedBuilder {
   final SearchResult searchResult;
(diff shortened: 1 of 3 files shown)


lib/src/search/search_navigation_controller.dart
```diff
@@ -104,14 +104,16 @@ class SearchNavigationController extends ValueNotifier<SearchNavigationState> {
...
+    final fullSelectionStart = codeController.code.hiddenRanges.recoverPosition(
+      visibleSelectionStart,
       placeHiddenRanges: TextAffinity.downstream,
     );
 
     var closestMatchIndex = _lastFullSearchResult.matches.indexWhere(
-      (element) => element.start >= fullSelectionEnd,
+      (element) {
+        return element.start >= fullSelectionStart;
+      },
```
alexeyinkin, line 116: Back to lambda?

test/src/search/search_navigation_test.dart
```diff
@@ -262,6 +262,41 @@ d
         null,
       );
     });
+
+    testWidgets('Changing search pattern with match at current '
+    'selection will not jump to next', (wt) async {
```
alexeyinkin, line 267: If you are testing that something is NOT happening, put the test to `issues` directory. Because most of the things the world can do to us do NOT happen, and it is unclear why we should test this one.

If keeping the test here, name it after what SHOULD happen.

lib/src/code_field/search_result_highlighted_builder.dart
```diff
@@ -4,7 +4,7 @@ import '../search/result.dart';
 import '../search/search_navigation_state.dart';
 
 @visibleForTesting
-const matchBackgroundColor = Color.fromARGB(141, 255, 235, 59);
+const matchBackgroundColor = Colors.green;
```
alexeyinkin, line 7: Let's use Chrome's colors instead as it is the most popular browser, and the people are used to it.
image

alexeyinkin (approved):

PR #254, opened 2023-08-19
Bumped http version to 1.1.0. Fixes #253

### LINKED ISSUE #253: use latest http pakage
as I want to use this package in my Flutter project but this package uses an older version of the HTTP package as other packages like google fonts use the latest package of HTTP so version mismatch occur.

5hirish:
This will also allow usage of google_fonts v4.0.5 +


nausharipov:
Thank you. Packages have been updated in https://github.com/akvelon/flutter-code-editor/pull/256.

PR #276, opened 2024-07-11
Autocomplete features

waseemdev:
Hi, 
Here are two commits:
1. I made `Autocompleter` abstract class and renamed default one to `DefaultAutocompleter`, so user can create its own implementation
    - also created `SuggestionItem` to be more customizable than just `String`, in some cases the displayed text in the popup list can be different than the actual code.

2. made the popup list customizable so user can use his own widget using `autocompleteListBuilder`.
   - also moved these variables (`caretPadding`, `autocompletePopupMaxHeight`, `autocompletePopupMaxWidth`) to be fields in `CodeEditor` widget

### DIFF
--- a/lib/src/autocomplete/autocompleter.dart
+++ b/lib/src/autocomplete/autocompleter.dart
@@ -1,133 +1,29 @@
-import 'package:autotrie/autotrie.dart';
-import 'package:highlight/highlight_core.dart';
+import 'package:flutter/material.dart';
+import 'package:highlight/highlight.dart';
 
-import '../code/reg_exp.dart';
+abstract class Autocompleter {
+  Mode? mode;
+  List<String> blacklist = [];
 
-/// Accumulates textual data and suggests autocompletion based on it.
-class Autocompleter {
-  Mode? _mode;
-  final _customAutocomplete = AutoComplete(engine: SortEngine.entriesOnly());
-  final _keywordsAutocomplete = AutoComplete(engine: SortEngine.entriesOnly());
-  final _textAutocompletes = <Object, AutoComplete>{};
-  final _lastTexts = <Object, String>{};
-  Set<String> _blacklistSet = const {};
+  Autocompleter();
 
-  static final _whitespacesRe = RegExp(r'\s+');
+  void setText(Object key, String? text);
 
-  /// The language to automatically extract keywords from.
-  Mode? get mode => _mode;
+  Future<List<SuggestionItem>> getSuggestionItems(TextEditingValue value);
 
-  set mode(Mode? value) {
-    _mode = value;
-    _parseKeywords();
-  }
+  TextEditingValue? replaceText(
+    TextSelection selection, TextEditingValue value, SuggestionItem item,
+  );
+}
 
-  void _parseKeywords() {
-    _keywordsAutocomplete.clearEntries();
+class SuggestionItem {
+  final String text;
+  final String displayText;
+  final dynamic data;
 
-    final keywords = mode?.keywords;
-    if (keywords == null) {
-      return;
-    }
-
-    if (keywords is String) {
-      _parseStringKeywords(keywords);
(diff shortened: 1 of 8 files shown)


lib/src/autocomplete/autocompleter.dart
```diff
@@ -1,133 +1,29 @@
-import 'package:autotrie/autotrie.dart';
-import 'package:highlight/highlight_core.dart';
+import 'package:flutter/material.dart';
+import 'package:highlight/highlight.dart';
 
-import '../code/reg_exp.dart';
+abstract class Autocompleter {
+  Mode? mode;
+  List<String> blacklist = [];
```
nausharipov, line 6: autocompleter.dart declares concrete fields instead of abstract accessors
Problem: Autocompleter currently has

Mode? mode;
List<String> blacklist = []; DefaultAutocompleter implements getters/setters for mode and blacklist (and uses @override). You cannot override a concrete field with getters/setters — this causes analyzer/compile errors.
Fix: make those members abstract (declare getters/setters) in Autocompleter. Example replacement:

Mode? get mode; set mode(Mode? value);

List<String> get blacklist; set blacklist(List<String> value);

(Keep the other abstract methods: setText, getSuggestionItems, replaceText.)


Example: 
```
import 'package:flutter/material.dart'; import 'package:highlight/highlight_core.dart';

abstract class Autocompleter { Autocompleter();

// Language mode used to extract keywords from highlight Mode Mode? get mode; set mode(Mode? value);

// Words to exclude from suggestions List<String> get blacklist; set blacklist(List<String> value);

/// Sets the [text] to parse all words from. /// Multiple texts are supported, each with its own [key]. /// Use this to set current texts from multiple controllers. void setText(Object key, String? text);

/// Returns items ready to be shown in popup based on current editing value Future<List<SuggestionItem>> getSuggestionItems(TextEditingValue value);

/// Replace selection/value using picked suggestion item. /// Returns a new TextEditingValue to assign to controller.value, or null if nothing changed. TextEditingValue? replaceText( TextSelection selection, TextEditingValue value, SuggestionItem item, ); }

class SuggestionItem { final String text; final String displayText; final dynamic data;
```

lib/src/autocomplete/default_autocompleter.dart
```diff
@@ -0,0 +1,210 @@
+import 'package:autotrie/autotrie.dart';
+import 'package:flutter/material.dart';
+import 'package:highlight/highlight_core.dart';
```
nausharipov, line 3: Autocompleter imports package:highlight/highlight.dart while DefaultAutocompleter imports package:highlight/highlight_core.dart. Choose the correct import that exposes Mode (typically highlight_core.dart) and use it in autocompleter.dart so Mode is known.

lib/src/autocomplete/default_autocompleter.dart
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    final result = await getSuggestions(prefix);
+
+    return result
+      .map((e) => SuggestionItem(text: e, displayText: e))
+      .toList();
+  }
+
+  Future<List<String>> getSuggestions(String prefix) async {
```
nausharipov, line 147: Decide whether getSuggestions(String) should be part of the Autocompleter interface (currently only DefaultAutocompleter exposes it; CodeController uses getSuggestionItems, so this is optional).

lib/src/autocomplete/autocompleter.dart
```diff
@@ -1,133 +1,29 @@
-import 'package:autotrie/autotrie.dart';
-import 'package:highlight/highlight_core.dart';
+import 'package:flutter/material.dart';
```
nausharipov, line 1: Since autocompleter.dart is a non-UI, library-level file (it only needs the types, not Material widgets), prefer the smaller import for TextEditingValue and TextSelection to avoid pulling Material into a low-level lib file: import 'package:flutter/services.dart';

PR #292, opened 2025-03-22
enhancement/Add FVM Support

Dario-Maselli:
## What type of PR is this?
- 🛠 Enhancement

## Description
Adds support for FVM. If a user has FVM installed, they just run `fvm use`.

## Added tests?
- 🙅 no, because they aren't needed

## Added to documentation?
- 🙅 No documentation needed


Dario-Maselli:
Re-opening shortly as I made a mistake with my branches 🤦‍♂️

PR #294, opened 2025-03-22
Enhancement/Add FVM Support

Dario-Maselli:
## What type of PR is this?
- 🛠 Enhancement

## Description
Adds support for FVM. If a user has FVM installed, they just run `fvm use`.

## Added tests?
- 🙅 no, because they aren't needed

## Added to documentation?
- 🙅 No documentation needed


.fvm/versions/3.29.2
```diff
@@ -0,0 +1 @@
+C:/Users/dario/fvm/versions/3.29.2
```
nausharipov, line 1: I doubt that FVM support is necessary for this project. It can easily be installed by users locally if they want it. Also, `C:/Users/dario/fvm/versions/3.29.2` path is only specific to your setup.

nausharipov:
@Dario-Maselli feel free to reply to my comment and reopen this PR if you don't agree.

PR #318, opened 2025-12-25
Support for Arabic Language and RTL Directionality

### LINKED ISSUE #317: Support for Arabic Language and RTL Directionality
Currently, the `flutter_code_editor` does not properly handle **Right-to-Left (RTL)** languages, specifically Arabic. This causes several UI and UX issues when writing or displaying Arabic code comments, strings, or documentation within the editor.

### **Current Behavior**

* Text alignment defaults to Left-to-Right (LTR) even when the system locale or specific text is Arabic.
* Cursors and selection highlights might behave unpredictably in RTL contexts.
* Punctuation and mixed LTR/RTL text (BiDi) are rendered incorrectly, breaking the code's visual structure.

### **Expected Behavior**

* The editor should detect or allow setting the `textDirection` to `TextDirection.rtl`.
* Proper cursor placement and movement logic for RTL text.
* Support for Arabic characters without overlapping or rendering issues in the code lines.

### **Proposed Solution (I can contribute this!)**

I have investigated the rendering logic and the `CodeField` implementation. I plan to submit a PR that:

1. Integrates `textDirection` support into the editor's styling.
2. Fixes the horizontal scrolling and alignment for RTL lines.
3. Ensures the gutter (line numbers) stays correctly positioned relative to the RTL text.

iskepr:
## What type of PR is this?

* 🍕 Feature
* 🛠 Enhancement

## Description

This PR adds support for **Arabic language and RTL directionality** in the editor.

* Implemented `textDirection` handling to allow proper rendering of RTL scripts.
* Ensured that the line numbers (gutter) maintain their correct alignment regardless of text direction.
* Closes #317.

## Added tests?

* 🙅 no, because they aren't needed

## Added to documentation?

* 🙅 No documentation needed

### DIFF
--- a/lib/src/code/reg_exp.dart
+++ b/lib/src/code/reg_exp.dart
@@ -1,3 +1,4 @@
 class RegExps {
-  static final wordSplit = RegExp('[^_a-zA-Z0-9]+');
+  static final wordSplit =
+      RegExp(r'[^_a-zA-Z0-9\u0600-\u06FF\u0750-\u077F\u08A0-\u08FF]+');
 }
--- a/lib/src/code_field/code_field.dart
+++ b/lib/src/code_field/code_field.dart
@@ -173,6 +173,7 @@ class CodeField extends StatefulWidget {
   final Color? background;
   final EdgeInsets padding;
   final Decoration? decoration;
+  final TextDirection? textDirection;
   final TextSelectionThemeData? textSelectionTheme;
   final FocusNode? focusNode;
 
@@ -196,6 +197,7 @@ class CodeField extends StatefulWidget {
     this.smartQuotesType = SmartQuotesType.disabled,
     this.padding = EdgeInsets.zero,
     GutterStyle? gutterStyle,
+    this.textDirection,
     this.enabled,
     this.readOnly = false,
     this.cursorColor,
@@ -234,7 +236,7 @@ class _CodeFieldState extends State<CodeField> {
 
   FocusNode? _focusNode;
   String? lines;
-  String longestLine = '';
+  var longestLine = '';
   Size? windowSize;
   late TextStyle textStyle;
   Color? _backgroundCol;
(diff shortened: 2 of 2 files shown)


lib/src/code_field/code_field.dart
```diff
@@ -196,6 +197,7 @@ class CodeField extends StatefulWidget {
     this.smartQuotesType = SmartQuotesType.disabled,
     this.padding = EdgeInsets.zero,
     GutterStyle? gutterStyle,
+    this.textDirection = TextDirection.ltr,
```
nausharipov: Defaulting textDirection to TextDirection.ltr causes a behavior regression for apps relying on inherited Directionality (RTL locales still render LTR unless every call site opts in). Please make this nullable and fall back to ambient direction, or default from Directionality.of(context).

lib/src/code_field/code_field.dart
```diff
@@ -464,17 +467,19 @@ class _CodeFieldState extends State<CodeField> {
...
-            if (widget.gutterStyle.showGutter) _buildGutter(),
-            Expanded(key: _editorKey, child: editingField),
-          ],
+      child: Directionality(
+        textDirection: widget.textDirection,
+        child: Container(
+          decoration: widget.decoration,
+          color: _backgroundCol,
+          key: _codeFieldKey,
+          child: Row(
```
nausharipov, line 477: This change drops the existing 8px editor inset. Was that intentional? It alters baseline layout for LTR and RTL. If padding needs to be direction-aware, consider EdgeInsetsDirectional.only(start: 8) instead of removing it.

lib/src/code/reg_exp.dart
```diff
@@ -1,3 +1,3 @@
 class RegExps {
-  static final wordSplit = RegExp('[^_a-zA-Z0-9]+');
+  static final wordSplit = RegExp(r'[^_a-zA-Z0-9\u0600-\u06FF]+');
```
nausharipov: now includes only \u0600-\u06FF for Arabic script. Is this intentional scope? Arabic text can also include characters from extended blocks (e.g. \u0750-\u077F, \u08A0-\u08FF) depending on language/content.

lib/src/code_field/code_field.dart
iskepr: done

lib/src/code/reg_exp.dart
iskepr: done

PR #319, opened 2026-03-03
Fix Chinese pinyin input issues in the editor

### LINKED ISSUE #126: [Windows Descktop] IME doesn't work well
IME: Microsoft [MSPY].
flutter_code_editor verson: 0.1.13

When type something in IME Chinese mode, unexpectation happened
https://user-images.githubusercontent.com/15630211/207526300-570bb3f2-cc78-4870-aa30-a3ab43d7011f.mp4

FeodorFitsner:
This PR fixes Chinese pinyin input issues in the editor.

Fixes #126

Root causes:
1) composing-only updates were ignored in CodeController.value
2) editor transforms could mutate IME in-progress composition state
3) custom span rendering bypassed Flutter’s composing-aware rendering
4) edit diffing could fail on transient out-of-range selections from IME

Changes:
- Make CodeController accept composing-only updates
- While composing is active, apply platform value updates directly
- Use super.buildTextSpan during active composition
- Clamp old/new selections in Code.getEditResult before diffing
- Prevent Enter/Tab/Esc shortcut actions during active composition

Tests:
- Added/updated tests for composing updates, IME commit path, and shortcut behavior
- Targeted test suite passes

### DIFF
--- a/lib/src/code_field/code_controller.dart
+++ b/lib/src/code_field/code_controller.dart
@@ -462,6 +462,7 @@ class CodeController extends TextEditingController {
 
   @override
   set value(TextEditingValue newValue) {
+    final hadActiveCompositionInOldValue = hasActiveComposition;
     final hasTextChanged = newValue.text != super.value.text;
     final hasSelectionChanged = newValue.selection != super.value.selection;
     final hasComposingChanged = newValue.composing != super.value.composing;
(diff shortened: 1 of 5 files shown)


lib/src/code_field/code_controller.dart
```diff
@@ -173,6 +173,8 @@ class CodeController extends TextEditingController {
     _code = _createCode(text ?? '');
     fullText = text ?? '';
 
+    debugPrint('CodeController initialized with text:\n$text');
```
nausharipov, line 176: debugPrint('CodeController initialized with text:\\n$text'); logs full editor content on every controller initialization. This can leak sensitive user code/data and create noisy logs for large documents. Please remove this log (or guard it behind an explicit opt-in debug flag that defaults off).

Issue #312, opened 2025-09-22
Gutter Numbers and Code Text Not Aligned


simeonangelov94:
@rohansingh @UbayGD @AKosolapov @mrshakirov @ArinFaraj 

Hello, since the package is now updated, I would kindly ask is it possible to add Gutter padding.

Currently, the gutter numbers and the code editor are not aligned. Please find example code below. If gutterPadding is added which I did before in a forked library - it aligns perfectly but otherwise is 4 units misaligned. Please find a photo and code example below.

Example Code
@override
  Widget build(BuildContext context) {
    //controller.visibleSectionNames = {'section1'};
    return CodeTheme(
      data: CodeThemeData(styles: customMonokaiSublimeTheme),
      child: CodeField(
        controller: codeEditorController,
        onChanged: (value) {
          if (widget.onChanged != null) widget.onChanged!(value);
        },
        padding: EdgeInsets.all(0),
        // gutterPadding: 4,
        minLines: widget.isBigScreen == false ? 5 : 15,
        // maxLines: widget.isBigScreen == false ? 15 : 25,
        textStyle:
            TextStyle(fontFamily: 'SourceCode', height: 1.5, fontSize: 18),
        gutterStyle: GutterStyle(
          margin: 10,
          textAlign: TextAlign.end,
          textStyle: const TextStyle(
              color: Colors.blue,
              fontWeight: FontWeight.bold,
              height: 1.5,
              fontFamily: 'SourceCode',
              fontSize: 18
              //fontSize: 18.0,
              ),
        ),
      ),
    );
  }

<img width="246" height="370" alt="Image" src="https://github.com/user-attachments/assets/7bbd7e5c-b4d1-454a-b377-8c35c096379f" />

Adityadgithub:
Try this - 
   Expanded(
  child: CodeTheme(
    data: CodeThemeData(styles: ThemeService.getTheme(_selectedTheme)),
    child: CodeField(
      gutterStyle: GutterStyle(
        showLineNumbers: true,
        showErrors: true,
        background: Colors.transparent,
        // FIX 1: Increased width to accommodate more digits (e.g., up to 999).
        width: 80, 
        textStyle: GoogleFonts.firaCode(
          fontSize: 12,
          color: const Color(0xFF00FF88),
          fontWeight: FontWeight.w500,
          // FIX 2: Added matching height property for vertical alignment.
          height: 1.6, 
        ),
[...]

PR #320, opened 2026-03-03
Fix gutter line-number alignment with editor text

### LINKED ISSUE #215: Gutter and text is not vertically aligned
I think, the gutter is off by one pixel, see:
image

Here's my configuration:
```dart
          CodeField(
            textStyle: const TextStyle(
              fontFamily: 'Cascadia Mono',
              // height: 1.5, // this seems to be the default
              fontSize: 16,
            ),
            gutterStyle: const GutterStyle(
              textStyle: TextStyle(
                color: Colors.white24,
                fontSize: 16,
                height: 1.5, // without this line, it doesn't match the text
              ),
            ),
            ...
```

I'm using version 0.2.15.

FeodorFitsner:
Fixes vertical misalignment between gutter line numbers and code text in `CodeField`.

Fixes: #215, #312 

Gutter and editor were using different/duplicated vertical spacing paths, which caused numbers to drift from code lines.

## Changes

- Unified padding so gutter and editor share the same vertical origin.
- Removed duplicated vertical padding paths.
- Kept line-height behavior consistent between code and line numbers.
- Adjusted popup offset calculations to match the new padding model.

## Files changed
- `lib/src/code_field/code_field.dart`
- `lib/src/gutter/gutter.dart`

## Validation
- Checked visually in example app (`03.change_language_theme`).
- Ran:
  - `dart analyze lib/src/code_field/code_field.dart lib/src/gutter/gutter.dart`
  - `flutter test test/src/code_field/code_controller_shortcut_test.dart`

### DIFF
--- a/lib/src/code_field/code_field.dart
+++ b/lib/src/code_field/code_field.dart
@@ -234,7 +234,7 @@ class _CodeFieldState extends State<CodeField> {
 
   FocusNode? _focusNode;
   String? lines;
-  String longestLine = '';
+  var longestLine = '';
   Size? windowSize;
   late TextStyle textStyle;
   Color? _backgroundCol;
@@ -390,9 +390,7 @@ class _CodeFieldState extends State<CodeField> {
     );
 
     return SingleChildScrollView(
-      padding: EdgeInsets.only(
-        right: widget.padding.right,
-      ),
+      padding: EdgeInsets.zero,
       scrollDirection: Axis.horizontal,
       controller: _horizontalCodeScroll,
       child: intrinsic,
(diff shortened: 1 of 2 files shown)


lib/src/code_field/code_field.dart
```diff
@@ -550,7 +556,6 @@ class _CodeFieldState extends State<CodeField> {
   double _getPopupLeftOffset(TextPainter textPainter) {
     return max(
       _getCaretOffset(textPainter).dx +
-          widget.padding.left -
```
nausharipov, line 553: horizontal scroll offset sign looks inverted; popup should stay anchored to caret while scrolling, so this likely needs subtracting scroll offset (not adding).

lib/src/code_field/code_field.dart
```diff
@@ -561,8 +566,6 @@ class _CodeFieldState extends State<CodeField> {
     return max(
       _getCaretOffset(textPainter).dy +
           caretHeight +
-          16 +
```
nausharipov, line 564: vertical scroll offset sign also appears inverted; with current math popup will drift when editor scrolls.

lib/src/gutter/gutter.dart
nausharipov, line 1: please add widget tests covering (1) line-number/text baseline alignment with non-default TextStyle.height, and (2) autocomplete popup position after horizontal/vertical scroll, since this PR changes both text metrics and popup offset math.

